PR #12, opened 2023-12-20
implementing Codespell into redo style build process

Jbsco:
Reopening #11

### DIFF
--- a/redo/rules/build_style.py
+++ b/redo/rules/build_style.py
@@ -95,7 +95,8 @@ class build_style(build_rule_base):
                           + " -S "
                           + os.path.basename(style_log_file)
                           + " 2>&1 | tee -a "
-                          + style_log_file)
+                          + style_log_file
+                          + " 1>&2")
         shell.run_command("codespell "
                           + build_directory
                           + " -I "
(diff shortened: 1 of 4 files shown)


redo/rules/build_style.py
```diff
@@ -85,6 +85,28 @@ def _build(self, redo_1, redo_2, redo_3):
...
+                          + " 1>&2")
+        shell.run_command("codespell "
+                          + build_directory
+                          + " -I "
+                          + codespell_ignore
+                          + " -S "
+                          + os.path.basename(style_log_file)
+                          + " 2>&1 | tee -a "
+                          + style_log_file
+                          + " 1>&2")
```
dinkelk, line 108: Maybe try to combine some of the code duplication here to make maintenance easier.

Jbsco:
> I think this is not quite right. I think the directory you want to run this on is `directory` which is set near the top of the class. The current directory is usually just the repo root. What you want is the directory that the user ran `redo` from which is passed by the `redo_1`, etc. args.

Addressed in [a5fe26e](https://github.com/lasp/adamant/pull/12/commits/a5fe26e42536fd8e9e81fa279441414561c35980).

> Maybe try to combine some of the code duplication here to make maintenance easier.

Also addressed in [a5fe26e](https://github.com/lasp/adamant/pull/12/commits/a5fe26e42536fd8e9e81fa279441414561c35980).

dinkelk (approved):

PR #33, opened 2024-03-27
Assembly Submodel Dependency Fix

dinkelk:
fixing dependency handling for product packetizer model which would cause it to not invalidate the model cache when it should


dinkelk:
Other components to look at for similar issue: ccsds_router, fault_correction, parameters, task watchdog. The other two already inherit from assembly_submodel, the first two do not, but likely should.

PR #39, opened 2024-05-24
Fix typos

MyApaulogies:
I fixed some typos based on the output of the `style_all` Github action. Also, there are a few typos involving Python string concatenation I fixed that need to be checked -- I could tell that I fixed the output of `redo targets`, but I didn't check the output of `gen/templates`. I just `rg`ed for the type of mistake (i.e. `cd ~/adamant && vim $(rg '"\s+"' -l)`) and fixed it where applicable.


gen/templates/record/name.py
```diff
@@ -40,11 +40,11 @@ def __init__(
         if {{ field.name }} is not None:
 {% if field.is_enum %}
             assert isinstance({{ field.name }}, {{ field.type_model.name }}), \
-                ("Expected type for field '{{ field.name }}' to be '{{ field.type_model.name }}'"
+                ("Expected type for field '{{ field.name }}' to be '{{ field.type_model.name }}' "
                  " and instead found '" + str(type({{ field.name }})))
```
dinkelk, line 44: This would create two spaces.

dinkelk (approved):

PR #40, opened 2024-05-28
Update CONTRIBUTING.md to refine PR workflow and expectations

dinkelk:
Adding to CONTRIBUTING.md guidance on how PRs should be handled by contributors and maintainers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,11 +8,25 @@ While we encourage all contributions, remember that Adamant is used everyday to 
 
 ## Creating a Submission
 
-Adamant follows the [GitHub flow](https://docs.github.com/en/get-started/quickstart/github-flow) development process. You should begin by [forking](https://docs.github.com/en/get-started/quickstart/fork-a-repo) this repository. When you have finished making changes on your fork, create a pull request back to this repository.
+Adamant follows the [GitHub flow](https://docs.github.com/en/get-started/quickstart/github-flow) development process. You should begin by [forking](https://docs.github.com/en/get-started/quickstart/fork-a-repo) this repository. When you have finished making changes on your fork, create a [pull request](https://github.com/lasp/adamant/pulls) back to this repository.
 
-Please try to keep submissions small and focused on single fixes, enhancements, or additions. This makes changes easier to review and faster to merge.
+### Pull Request
 
-Once a pull request has been submitted the following will occur:
+Please adhere to the following when submitting a pull request:
+
+- Pull requests should be made against `main`.
+- Keep submissions small and focused on single fixes, enhancements, or additions. This makes changes easier to review and faster to merge.
+- Include a summary of changes that gives reviewers an idea of what they should pay attention to.
+- Pull request branches should have as "clean" of a history as possible.
+    - Each commit should present one change or idea to a reviewer.
+    - Commits that merely "fix up" previous commits should be interactively rebased and squashed into their targets.
+- Prefer the use of `git rebase`.
+    1.  `git rebase` actually _rebases_ your branch from the current main branch's endpoint. This localizes conflicts to the commits at which they actually appear, though it can become complicated when there are more than a few conflicts.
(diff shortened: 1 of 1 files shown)


Jbsco:
This looks effective, particularly emphasis on clean history and presenting changes clearly.

CONTRIBUTING.md
```diff
@@ -8,11 +8,25 @@ While we encourage all contributions, remember that Adamant is used everyday to
...
+Adamant follows the [GitHub flow](https://docs.github.com/en/get-started/quickstart/github-flow) development process. You should begin by [forking](https://docs.github.com/en/get-started/quickstart/fork-a-repo) this repository. When you have finished making changes on your fork, create a [pull request](https://github.com/lasp/adamant/pulls) back to this repository.
 
-Please try to keep submissions small and focused on single fixes, enhancements, or additions. This makes changes easier to review and faster to merge.
+### Pull Request
 
-Once a pull request has been submitted the following will occur:
+Please adhere to the following when submitting a pull request:
+
+- Pull requests should be made against `main`.
+- Keep submissions small and focused on single fixes, enhancements, or additions. This makes changes easier to review and faster to merge.
```
Jbsco, line 18: It may be less common but is it worth mentioning larger feature cases?

CONTRIBUTING.md
```diff
@@ -8,11 +8,25 @@ While we encourage all contributions, remember that Adamant is used everyday to
...
-Adamant follows the [GitHub flow](https://docs.github.com/en/get-started/quickstart/github-flow) development process. You should begin by [forking](https://docs.github.com/en/get-started/quickstart/fork-a-repo) this repository. When you have finished making changes on your fork, create a pull request back to this repository.
+Adamant follows the [GitHub flow](https://docs.github.com/en/get-started/quickstart/github-flow) development process. You should begin by [forking](https://docs.github.com/en/get-started/quickstart/fork-a-repo) this repository. When you have finished making changes on your fork, create a [pull request](https://github.com/lasp/adamant/pulls) back to this repository.
 
-Please try to keep submissions small and focused on single fixes, enhancements, or additions. This makes changes easier to review and faster to merge.
+### Pull Request
 
-Once a pull request has been submitted the following will occur:
+Please adhere to the following when submitting a pull request:
+
+- Pull requests should be made against `main`.
```
Jbsco, line 17: Is there a process that works for setting up multiple contributors on a single PR and is this worth mentioning here?

CONTRIBUTING.md
dinkelk: My thoughts are to encourage small feature updates. Obviously that won't always be possible, and those cases will have to be managed separately. But those should be the exception, not the rule. I question whether that needs to be addressed directly or not, in light of that.

CONTRIBUTING.md
dinkelk: Similar to my other comment. Since that is the exception, not the rule, it is hard to address with a single process IMO. But you are right, rebasing is problematic when more than one person is committing to a PR.

MyApaulogies:
I think what's been added makes sense. Looking at [another open source project](https://github.com/zellij-org/zellij/blob/main/CONTRIBUTING.md) for reference, I have a few thoughts thought, not necessarily to do with the specific changes you made.
1. "If this is not a trivial fix, consider creating an issue to discuss first and later link to it from the PR." I think this would be a good policy, to make development more visible.
2. Moving the testing section to the top, to match the flow of development
3. Elaborating on the testing section -- I think mentioning what the Github Actions do, and how you can run them on your local machine, would be helpful too. Something worth mentioning might be how if your changes are confined to a directory, you generally only have to worry about tests/style checks passing within that directory. Maybe it would be enough to link to a relevant section in the user guide?

dinkelk:
@MyApaulogies - I like those suggestions. I will try to incorporate.

dinkelk:
Added changes to address @MyApaulogies suggestions.

MyApaulogies (approved):

PR #50, opened 2024-06-19
Docker Compose Conversion

Jbsco:
These changes convert the current Docker scripts into a unified Docker Compose script which manages Docker container operations with arguments such as start, stop, login, push, build, and remove. The Docker configuration is contained within the compose.yaml file and is provided as an argument to the docker-compose command contained within the unified script.


docker/adamant_env.sh
```diff
@@ -0,0 +1,84 @@
...
+    }
+  else
+    echo "Neither docker nor podman found!!!"
+    exit 1
+  fi
+fi
+
+DOCKER_CONTAINER_NAME="adamant_container"
+DOCKER_IMAGE_NAME="ghcr.io/lasp/adamant:latest"
+DOCKER_COMPOSE_COMMAND="docker compose"
```
dinkelk, line 20: I am thinking we should still grab these from docker_config.sh. We shouldn't expect the user to ever look at this file, but may want to modify docker_config.sh.

docker/adamant_env.sh
```diff
@@ -0,0 +1,84 @@
...
+  exit 1
+}
+
+if [ "$#" -eq 0 ]; then
+  usage $0
+fi
+
+case $1 in
+  start )
+    execute "${DOCKER_COMPOSE_COMMAND} -f compose.yml up -d"
```
dinkelk, line 54: I am thinking that this may not run from anywhere with this implementation. For instance, can I run adamant_env.sh from a different directory, than this directory, and still have it work? You may need to use the BASH readlink trick to get `$this` to make this work.

dinkelk (changes requested):

docker/adamant_env.sh
Jbsco: Added in [243c4cc](https://github.com/lasp/adamant/pull/50/commits/243c4cc778370ac756b75879a5e6f0c304b2890d).

docker/adamant_env.sh
```diff
@@ -0,0 +1,87 @@
...
+  start )
+    execute "${DOCKER_COMPOSE_COMMAND} -f ${DOCKER_COMPOSE_CONFIG} up -d"
+    echo ""
+    echo "Run \"./adamant_env.sh login\" to log in."
+    ;;
+  stop )
+    execute "${DOCKER_COMPOSE_COMMAND} -f ${DOCKER_COMPOSE_CONFIG} stop"
+    ;;
+  login )
+    execute "docker exec -it -u user ${DOCKER_CONTAINER_NAME} //bin//bash"
```
dinkelk, line 65: shouldn't this just use `docker compose run`

docker/adamant_env.sh
```diff
@@ -0,0 +1,87 @@
...
+    ;;
+  push )
+    execute "docker push ${DOCKER_IMAGE_NAME}"
```
dinkelk, line 68: shouldn't this just use `docker compose push`

docker/adamant_env.sh
```diff
@@ -0,0 +1,87 @@
...
+    ;;
+  build )
+    execute "docker build --progress=plain -t ${DOCKER_IMAGE_NAME} -f Dockerfile ."
```
dinkelk, line 71: shouldn't this just use `docker compose build`?

dinkelk (changes requested):

docker/adamant_env.sh
Jbsco: Addressed in [37ca6e0](https://github.com/lasp/adamant/pull/50/commits/37ca6e0595b321efcb7d41225d9391152126db30).

docker/adamant_env.sh
Jbsco: Addressed in [cdaf5fe](https://github.com/lasp/adamant/pull/50/commits/cdaf5fed1b1dc48981e008a9a621e2d8810fea5c).

docker/adamant_env.sh
Jbsco: Addressed in [cdaf5fe](https://github.com/lasp/adamant/pull/50/commits/cdaf5fed1b1dc48981e008a9a621e2d8810fea5c).

docker/adamant_env.sh
Jbsco: Addressed in [cdaf5fe](https://github.com/lasp/adamant/pull/50/commits/cdaf5fed1b1dc48981e008a9a621e2d8810fea5c).

dinkelk (approved):

PR #51, opened 2024-06-21
Move Python documentation comments to docstring

MyApaulogies:
Automated with a script, but manually reviewed as well. Anywhere there is a relevant comment above a `def`/`class`, I moved it into a docstring. It's quite messy since it edits so many files, but it passed the tests.  ~~Also, here is the script~~


dinkelk:
I am looking at this: https://peps.python.org/pep-0257/

Is there a reason for choosing `'''` over `"""`. That link prefers the latter. I'd like to stick to whatever is the standard convention.

dinkelk (approved):

PR #52, opened 2024-06-27
`not_null` option in component init and discriminant parameters

MyApaulogies:
Note that there might be other places where it would make sense to add `not_null`, likely wherever the `parameter` submodule gets instantiated in a Python file in gen/models.

### DIFF
--- a/gen/models/submodels/variable.py
+++ b/gen/models/submodels/variable.py
@@ -64,7 +64,7 @@ class datatype(object):
 
 class variable(object):
     """Class holding information related to an Ada variable"""
-    def __init__(self, name, type, description=None, value=None, default_value=None):
+    def __init__(self, name, type, description=None, value=None, default_value=None, not_null=False):
         self.name = ada.formatVariable(name)
         self.datatype = datatype(name=type)
         self.description = description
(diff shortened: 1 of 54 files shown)


gen/models/submodels/variable.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, name, description=None):
 
 class variable(object):
     """Class holding information related to an Ada variable"""
-    def __init__(self, name, type, description=None, value=None, default_value=None):
+    def __init__(self, name, type, description=None, value=None, default_value=None, not_null=False):
```
dinkelk, line 67: I don't think we want to modify `variable.py` at all. This file represents an Ada variable, and all the things associated with it. `not null` is not a thing for variables. It is only allowed for parameters (ie. variables passed to subprograms). So I would expect that `self.not_null` only lives in `parameter.py` right below `self.mode`.

dinkelk (changes requested):

gen/models/submodels/variable.py
MyApaulogies: I think it is possible though, try adding `Y : not null Ref_Element := X;` to one of the .adb files here then compiling: https://learn.adacore.com/courses/advanced-ada/parts/resource_management/access_types.html#null-not-null-access

dinkelk: Wow. You are right. That is cool!

dinkelk (approved):

PR #53, opened 2024-06-27
Moving discriminant parameters to component base

dinkelk:
By moving the discriminant to the base type we ensure that the component model always matches the implementation. When the discriminant is reflected in the implementation, it was up to the user to keep the model in sync with the code, which is far from ideal. This commit fixes that issue.

### DIFF
--- a/gen/models/component.py
+++ b/gen/models/component.py
@@ -472,11 +472,16 @@ class component(base):
                 + (self.includes)
             )
         )
-        self.base_ads_includes = [
-            inc
-            for inc in self.base_ads_includes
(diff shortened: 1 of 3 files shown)


dinkelk:
Abandoning. This does not quite work with the way Ada handles inherited discriminants, ie.

```
error: inherited discriminant not allowed here (RM 3.8 (12), 3.8.1 (6))
```

dinkelk:
Abandoning. This does not quite work with the way Ada handles inherited discriminants, ie.

```
error: inherited discriminant not allowed here (RM 3.8 (12), 3.8.1 (6))
```

PR #56, opened 2024-07-03
Add Ada 2022 support for all build targets

dinkelk:


Jbsco (approved):
Tested on Linux with the example assembly, `redo` builds correctly, Pico flashes correctly, COSMOS picks it up with the serial plugin as normal.

PR #57, opened 2024-07-03
`activate` now always rebuilds `alr` dependencies.

dinkelk:
If a user did not remove their `alire/` directory when rebuilding their container environment, then the `alr` dependencies would not be downloaded and compiled correctly. This change ensures that every time a user logs in, the `alr` dependencies are rebuilt.


Jbsco (approved):
Tested on Linux with example assembly. Alire directory rebuilds each time on `adamant_env.sh login`, taking less than 1 second after the initial login, and allows using correct configuration in the assembly.

PR #68, opened 2024-08-13
Add a new basic assertion for `system.address`

Jbsco:
This pull request adds a new basic assertion for `system.address` types when unit testing. If merged, this would allow testing for memory address equality.


dinkelk (approved):
Good addition.

PR #80, opened 2024-09-17
Add Ada 2022 @ syntax

MyApaulogies:


dinkelk:
@MyApaulogies - can you use `rebase -i` to drop the first and last commits in the PR. We don't need commits where you add and then later remove the fsh script. Although it is pretty cool to see!

dinkelk:
@MyApaulogies's script used for replacement `at-syntax.fish`:
```
# requires moreutils and ripgrep
for i in 1 2
    set -l regex '^(\s*(\S.*?)\h+:=.*?(\h|\())\2((\.|\h|\)).*?;)'
    for file in (rg -P --multiline $regex -l)
        rg -P --multiline $regex -r '$1@$4' "$file" -N --passthru | sponge "$file"
    end
end
```

dinkelk (approved):

PR #91, opened 2025-01-07
COSMOS Plugin Telemetry Fix

Jbsco:
When generating the telemetry configuration for the COSMOS plugin, block data was only generated if the packet name was Events_Packet. Instead, the generator template now checks the packet item array, and produces the block data item in the correct case, similar to the Hydra template. Note that the CRC item is handled differently when the block data item is present, as the subpacket data has an indeterminate length at plugin compile time.


dinkelk:
Tested this on my system and it works great. Thank you!

PR #92, opened 2025-01-08
Fix for unit test log statement compiler error

dinkelk:


Jbsco (approved):
Tested with a component in test that was previously producing `error:   ==> in call to "To_Tuple_String"` errors in autocode relating to CCSDS packet receive functions, and found that this change resolves the issue.

Jbsco (approved):

PR #99, opened 2025-02-08
Update data dependencies generator template to apply `_Stale_Limit`

Jbsco:
When setting the ID and limits for a data product `_Stale_Limit` did not appear to be set. This presents as the default stale limit being used even when setting `map_data_dependencies: stale_limit_us:` in an assembly, and selecting `time: packet_time` in an extracted product.


dinkelk (approved):
Nice find!

PR #100, opened 2025-03-11
Build system performance enhancements

dinkelk:
For `redo clean` as well as building binaries and executables.

For `redo clean` we now call out to the system to remove files using `rm` and we do this concurrently. This is much faster than using python's `rmtree`.

For building, the prebuild logic was making duplicate calls out to `redo-ifchange` during the prebuild phase. A refactor was able to eliminate these duplicate calls, improving build times.


Jbsco (approved):
Performance appears much improved in build tests on Linux.

PR #105, opened 2025-05-15
Add Dump_Registers command to Register Stuffer

owhite7128:
This adds a command to enable multiple registers to be read in a single command through the register stuffer component.


src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +268,60 @@ package body Component.Register_Stuffer.Implementation is
...
+            return Failure;
+         end if;
+
+         for Register in 0 .. Arg.Value - 1 loop
+            if not Self.Is_Address_Valid (Arg.Address + Storage_Offset (Register * 4)) then
+               return Failure;
+            end if;
+            declare
+               -- Define the register at the appropriate address:
+               Reg : Packed_U32.Register_T_Le with Import, Convention => Ada, Address => Arg.Address + Storage_Offset (Register * 4);
```
dinkelk, line 297: Instead of hard coding 4, better to use Packed_U32.Size_In_Bytes.

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +268,60 @@ package body Component.Register_Stuffer.Implementation is
...
+               -- Read the register value:
+               Reg_Copy : constant Packed_U32.T := Packed_U32.Swap_Endianness (Packed_U32.T_Le (Reg));
+            begin
+               -- Load the register value, in Reg_Copy to the Packet type.
+               Arr (Register) := Reg_Copy;
```
dinkelk, line 302: Can just move the right hand side of 299 here.

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +268,60 @@ package body Component.Register_Stuffer.Implementation is
...
+            end;
+         end loop;
+         -- Construct and Send the packet based upon the returned array of register values
+         Self.Packet_T_Send_If_Connected (Self.Packets.Register_Packet_Truncate (The_Time,
```
dinkelk, line 306: Hm. Should never need to "trucate" since you types are defined to never allow that to occur.

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +268,60 @@ package body Component.Register_Stuffer.Implementation is
...
+            Register_Stuffer_Packet.Pack (Register_Stuffer_Packet.U'(
+               Header => Register_Stuffer_Packet_Header.U'(
+                  Start_Address => Arg.Address,
+                  Num_Registers => Arg.Value
+               ),
+               Buffer => Register_Stuffer_Packet_Array.Unpack (Arr)
```
dinkelk, line 312: We should be able to avoid all packing/unpacking here. None of this is required in this case.

dinkelk (changes requested):

src/components/register_stuffer/test/register_stuffer_tests-implementation.adb
```diff
@@ -391,4 +412,160 @@ package body Register_Stuffer_Tests.Implementation is
...
+   end Test_Nominal_Dump_One_Registers;
+
+   -- This unit test makes sure the component can dump the maximum number of
+   -- registers by command.
+   overriding procedure Test_Nominal_Dump_Max_Registers (Self : in out Instance) is
+      T : Component.Register_Stuffer.Implementation.Tester.Instance_Access renames Self.Tester;
+      function Generate_Index_Value (I : Interfaces.Unsigned_32) return Interfaces.Unsigned_32 is
+      begin
+         return I;  -- Simply return the index
+      end Generate_Index_Value;
```
dinkelk, line 470: This seems unnecessary

src/components/register_stuffer/test/register_stuffer_tests-implementation.adb
```diff
@@ -391,4 +412,160 @@ package body Register_Stuffer_Tests.Implementation is
...
+
+      -- Create array using a loop and expression function ((packet_buffer_size - 20) / 8)
+      Packet_Header_Length : constant Natural := Register_Stuffer_Packet_Header.Serialization.Serialized_Length;
+      Packet_Max_Registers : constant Natural := (Configuration.Packet_Buffer_Size - Packet_Header_Length) / (Packet_Header_Length - 4);
+
+      Large_Index_Array : array (1 .. Packet_Max_Registers) of Interfaces.Unsigned_32 :=
+         [for I in 1 .. Packet_Max_Registers => Generate_Index_Value (Interfaces.Unsigned_32 (I))];
```
dinkelk, line 477: If you are using this as the memory region to dump, you need to use packed types, otherwise the endianess is dependent. So better to use Packed_U32.T_Le array here.

dinkelk (changes requested):

src/components/register_stuffer/test/register_stuffer.tests.yaml
```diff
@@ -11,3 +11,9 @@ tests:
     description: This unit test makes sure an malformed command is rejected.
   - name: Test_Protected_Register_Write
     description: This unit test makes sure the protected register write feature works as intended.
+  - name: Test_Nominal_Dump_One_Registers
+    description: This unit test makes sure the component can dump one register by command.
+  - name: Test_Nominal_Dump_Max_Registers
+    description: This unit test makes sure the component can dump the maximum number of registers by command.
```
dinkelk, line 17: In this test, check to make sure the packet you get out is indeed maximum length.

src/components/register_stuffer/types/64bit/n_registers.record.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
dinkelk, line 1: This is basically a duplicate of the register_dump_packet_header type. I'd propose deleting this one and the 32 bit version and making everything use the header type.

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +232,75 @@ package body Component.Register_Stuffer.Implementation is
...
+         Last_Addr : System.Address;
+      begin
+         -- Loop on Number of Registers:
+         -- Somehow get the underlying integer from num_registers
+         if not Is_End_Address_Valid (Arg) then
+            Self.Event_T_Send_If_Connected (Self.Events.Address_Range_Overflow (Self.Sys_Time_T_Get, Arg));
+            return Failure;
+         end if;
+
+         if not Self.Is_Address_Valid (Arg.Address) then
```
dinkelk, line 257: Let's do this check before the end address check. I think it makes most sense to check start address prior.

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +232,75 @@ package body Component.Register_Stuffer.Implementation is
...
+                  Header => (
+                     Start_Address => Arg.Address,
+                     Num_Registers => Arg.Value
+                  ),
+                  Buffer => Arr
+               ),
+               Packet_To_Send
+            );
+         begin
+            if Serial_Status = Failure then
```
dinkelk, line 285: Better to just assert here since this should happen. We will check this with unit test.
```
pragma Assert (Serial_Status = Success);
```

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +232,75 @@ package body Component.Register_Stuffer.Implementation is
...
+            return Failure;
+         end if;
+
+         for Register in 0 .. Arg.Value - 1 loop
+            declare
+               -- Define the register at the appropriate address:
+               Reg : Packed_U32.Register_T_Le with Import, Convention => Ada, Address => Arg.Address + Storage_Offset (Register * Packed_U32.Size_In_Bytes);
+            begin
+               -- Load the register value, in Reg_Copy to the Packet type.
+               Last_Addr := Arg.Address + Storage_Offset (Register * Packed_U32.Size_In_Bytes);
```
dinkelk, line 267: Would be better to just compute this once and not calculate in a loop. You can easily determine this using Arg.Address and Arg.Value and calculate the last address as a constant.

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +232,75 @@ package body Component.Register_Stuffer.Implementation is
...
+            end;
+         end loop;
+         -- Construct and Send the packet based upon the returned array of register values
+         declare
+            Packet_To_Send : Packet.T;
+            Serial_Status : constant Serialization_Status := Self.Packets.Register_Packet (The_Time,
+               (
+                  Header => (
+                     Start_Address => Arg.Address,
+                     Num_Registers => Arg.Value
```
dinkelk, line 278: Arg.Value is not a good name. This should be something like "length", although this might get resolved when you delete this type, see my comment further down.

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +232,75 @@ package body Component.Register_Stuffer.Implementation is
...
+      end if;
+
+      declare
+         Arr : Register_Dump_Packet_Array.T := [others => (Value => 0)];
+         Last_Addr : System.Address;
+      begin
+         -- Loop on Number of Registers:
+         -- Somehow get the underlying integer from num_registers
+         if not Is_End_Address_Valid (Arg) then
+            Self.Event_T_Send_If_Connected (Self.Events.Address_Range_Overflow (Self.Sys_Time_T_Get, Arg));
```
dinkelk, line 253: Use `The_Time` here.

src/components/register_stuffer/test/register_stuffer_tests-implementation.adb
```diff
@@ -391,4 +413,167 @@ package body Register_Stuffer_Tests.Implementation is
...
+   end Test_Nominal_Dump_One_Registers;
+
+   -- This unit test makes sure the component can dump the maximum number of
+   -- registers by command.
+   overriding procedure Test_Nominal_Dump_Max_Registers (Self : in out Instance) is
+      T : Component.Register_Stuffer.Implementation.Tester.Instance_Access renames Self.Tester;
+
+      -- Create array using a loop and expression function ((packet_buffer_size - 20) / 8)
+      Packet_Header_Length : constant Natural := Register_Dump_Packet_Header.Serialization.Serialized_Length;
+      Packet_Max_Registers : constant Natural := (Configuration.Packet_Buffer_Size - Packet_Header_Length) / (Packed_U32.Size_In_Bytes);
```
dinkelk, line 471: This is not a good way to do this. You should use the type max, ie `N_Registers. Num_Registers'Last`. You want this test to check that your configuration magic works, so do not duplicate the configuration magic here. You should not be withing Configuration in this unit test at all.

src/components/register_stuffer/test/register_stuffer_tests-implementation.adb
```diff
@@ -391,4 +413,167 @@ package body Register_Stuffer_Tests.Implementation is
       Packed_Arm_Timeout_Assert.Eq (T.Armed_State_Timeout_History.Get (9), (Timeout => 0));
    end Test_Protected_Register_Write;
 
+   -- This unit test makes sure the component can dump one register by command.
+   overriding procedure Test_Nominal_Dump_One_Registers (Self : in out Instance) is
+      T : Component.Register_Stuffer.Implementation.Tester.Instance_Access renames Self.Tester;
+      Packet_Header_Length : constant Natural := Register_Dump_Packet_Header.Serialization.Serialized_Length;
+   begin
+      T.Command_T_Send (T.Commands.Dump_Registers ((Address => Register_Array'Address, Value => 1)));
```
dinkelk, line 421: Instead of hard coding 1, do `'First` of the type. This helps you test that you defined the type appropriately.

dinkelk (changes requested):

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +232,67 @@ package body Component.Register_Stuffer.Implementation is
...
+            Packet_To_Send : Packet.T;
+            Serial_Status : constant Serialization_Status := Self.Packets.Register_Packet (The_Time,
+               (
+                  Header => Arg,
+                  Buffer => Arr
+               ),
+               Packet_To_Send
+            );
+         begin
+            pragma Assert (Serial_Status = Success);
```
dinkelk, line 279: You need to justify this assertion with a comment above it. Basically, this should never fail because the packet type is designed to never be larger than a single packet.

dinkelk (changes requested):

src/components/register_stuffer/component-register_stuffer-implementation.adb
```diff
@@ -225,6 +239,68 @@ package body Component.Register_Stuffer.Implementation is
...
+      declare
+         Arr : Register_Dump_Packet_Array.T := [others => (Value => 0)];
+         Last_Addr : constant System.Address := Arg.Start_Address + Storage_Offset ((Arg.Num_Registers - 1) * Packed_U32.Size_In_Bytes);
+      begin
+         if not Self.Is_Address_Valid (Arg.Start_Address) then
+            return Failure;
+         end if;
+
+         if not Is_End_Address_Valid (Arg) then
+            Self.Event_T_Send_If_Connected (Self.Events.Address_Range_Overflow (The_Time, Arg));
```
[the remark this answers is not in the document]
owhite7128: I agree, that makes sense

dinkelk (approved):

PR #106, opened 2025-06-03
Add optional Data Dependency arrows to show data flow across a view, …

owhite7128:
…also add filters to filter data dependencies and data products

### DIFF
--- a/doc/user_guide/user_guide.tex
+++ b/doc/user_guide/user_guide.tex
@@ -5656,6 +5656,7 @@ Following is a set of \texttt{show\_*} switches. By default, if left unspecified
   \item \textbf{\texttt{show\_component\_name}} - Shows the component instance name inside each component box.
   \item \textbf{\texttt{show\_connector\_type}} - Shows the connector type next to each connection.
   \item \textbf{\texttt{hide\_group\_outline}} - If set to \texttt{False}, it shows a dotted line around any ``groups" defined in the view
+  \item \textbf{\texttt{show\_data\_dependencies}} - This flag is an exception, it is default set to \texttt{False}. If set to \texttt{True}, any data dependencies defined in the assembly are displayed.
 \end{spaceditemize}
 \vspace{5mm} %5mm vertical space
 
@@ -5665,7 +5666,7 @@ Below the \texttt{rule} definition is the \texttt{filters}. Any number of filter
 
 In this case we define a single \texttt{filter}. A \texttt{name} must be provided for the filter, and it must be unique within the view. A filter \texttt{type} must also be provided. Along with the filter type is one of two lists (but never both): an \texttt{include} list or an \texttt{exclude} list. If an \texttt{include} list is provided, then the assembly graph is filtered to \textit{include} any components or connections that meet the filter criteria to create the subgraph. If an \texttt{exclude} list is provided, then the assembly graph is filtered to include any components or connections that not NOT meet the filter criteria to create the subgraph. \\
 
-There are two different types of view filters: 1) those that filter components, and 2) those that filter connectors. A component filter will remove any components that do not meet the filter criteria. Any connections that are shared between the components are left, and any connections that connect to removed components are pruned from the subgraph. The component filters are shown below:
+There are three different types of view filters: 1) those that filter components, 2) those that filter connectors, and 3) those that filter data dependencies. A component filter will remove any components that do not meet the filter criteria. Any connections that are shared between the components are left, and any connections that connect to removed components are pruned from the subgraph. The component filters are shown below:
 
 \vspace{5mm} %5mm vertical space
 \begin{spaceditemize}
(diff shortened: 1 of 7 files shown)


gen/schemas/view.yaml
```diff
@@ -38,6 +38,9 @@ mapping:
   hide_group_outline:
     type: bool
     required: False
+  hide_data_dependencies:
```
dinkelk, line 41: I am thinking this is better called "show_data_dependencies" and it defaults to False. From a user perspective, I'd like to think about it as enabling a feature to show them. Do you agree?

gen/models/view.py
```diff
@@ -97,6 +97,40 @@ def check_connector_types(filter_obj, assm):
...
+
+
+def check_data_dependency_types(filter_obj, assm):
+    data_dep_types = get_data_dependency_types(assm)
+    for t in filter_obj.item_list:
+        if t not in data_dep_types:
+            warn(
+                "In filter '"
+                + filter_obj.name
+                + "' connector type '"
```
dinkelk, line 111: not connector type. data dependency type

gen/models/view.py
```diff
@@ -97,6 +97,40 @@ def check_connector_types(filter_obj, assm):
             )
 
 
+def get_data_dependency_types(assm):
```
dinkelk, line 100: Isn't this returning the whole data dependency object? Not just the type of the data dependencies.

gen/models/view.py
```diff
@@ -97,6 +97,40 @@ def check_connector_types(filter_obj, assm):
...
+            warn(
+                "In filter '"
+                + filter_obj.name
+                + "' connector type '"
+                + str(t)
+                + "' does not exist in assembly."
+            )
+
+
+def get_data_dependency_names(assm):
```
dinkelk, line 117: Add some comments to these functions. Looks like you are appending list of dp names to dd names.

gen/models/view.py
```diff
@@ -97,6 +97,40 @@ def check_connector_types(filter_obj, assm):
...
+
+
+def check_data_dependency_names(filter_obj, assm):
+    data_dep_names = get_data_dependency_names(assm)
+    for t in filter_obj.item_list:
+        if t not in data_dep_names:
+            warn(
+                "In filter '"
+                + filter_obj.name
+                + "' connector type '"
```
dinkelk, line 129: fix, not connector type

gen/models/view.py
```diff
@@ -420,6 +454,66 @@ def f(assm_to_filter):
...
+
+def include_data_dependency_name_filter(filter_obj, assm):
+    check_data_dependency_names(filter_obj, assm)
+
+    def f(assm_to_filter):
+        data_deps = {}
+        for id, dd in assm_to_filter.data_dependencies.items():
+            if dd.suite.component.instance_name + "." + dd.name in filter_obj.item_list:
+                data_deps[id] = dd
+            elif dd.data_product.suite.component.instance_name + "." + dd.data_product.name in filter_obj.item_list:
```
dinkelk, line 493: Probably better to use an "or" instead of elif here. This is the same pattern copied from connector_names filter I see, so that should be fixed too.

gen/models/view.py
```diff
@@ -444,28 +538,45 @@ def f(assm_to_filter):
...
+    ("data_dependency_name", True): include_data_dependency_name_filter,
+    ("data_dependency_name", False): exclude_data_dependency_name_filter,
 }
 
 
 def prune(assm_to_filter):
     component_names = [x.instance_name for x in assm_to_filter.components.values()]
 
-    # Cleanup 1: Remove connections for which there is no component:
+    # Cleanup 1: Build a set of unfiltered components based on connectors:
```
dinkelk, line 551: The previous comment made more sense to me.

gen/models/view.py
```diff
@@ -444,28 +538,45 @@ def f(assm_to_filter):
...
         from_component = connection.from_component.instance_name
         if (to_component in component_names) and (from_component in component_names):
-            new_connections.append(connection)
             components_with_connections.append(to_component)
             components_with_connections.append(from_component)
+            new_connections.append(connection)
     assm_to_filter.connections = new_connections
 
-    # Cleanup 2: Remove components for which there is no connectors:
+    # Cleanup 2: Build a set of unfiltered components based on data dependencies:
```
dinkelk, line 563: Isn't this really, remove data dependencies for which there is no component?

gen/models/view.py
```diff
@@ -610,12 +746,14 @@ def __init__(self, name, type, item_list=[], list_type="exclude"):
...
         # Format lists correctly:
-        if self.type in ["component_type", "connector_type", "component_type_context"]:
+        if self.type in ["component_type", "connector_type", "component_type_context", "data_dependency_type"]:
             self.item_list = list(set(map(lambda x: ada.formatType(x), item_list)))
         elif self.type in [
             "component_name",
             "connector_name",
             "component_name_context",
+            "data_dependency_name",
+            "data_product_name",
```
dinkelk, line 756: I thought we were not doing "data_product_name". Maybe look and see if this got in here anywhere else.

gen/models/view.py
```diff
@@ -1005,4 +1145,7 @@ def new_assembly(data):
         new_data["subassemblies"] = []
 
         # Create an assembly out of the new data and return it:
-        return new_assembly(new_data)
+        ret_assm = new_assembly(new_data)
+        # Pass data_dependencies through to the new assembly
+        ret_assm.data_dependencies = new_assm.data_dependencies
```
dinkelk, line 1150: Can this not be done within the `new_assembly` function?

dinkelk (changes requested):

doc/user_guide/user_guide.tex
```diff
@@ -5689,6 +5690,15 @@ \subsubsection{Creating a View Model}
 
 In this case we use the \texttt{connector\_type} filter type with an \texttt{include} list. In this way we are trying to include any connectors that are of type \texttt{Command.T}. Any components that do not have a \texttt{Command.T} connector will be removed. \\
 
+Data dependency filters should always be used in conjunction with \textbf{\texttt{show\_data\_dependencies}} flag set to \texttt{True}. These filters operate exclusively with connector filters, so data dependency filters will leave all connectors unfilterd, and vice versa.
```
dinkelk, line 5693: What do you mean by exclusively? Do you mean "independently of"?

PR #120, opened 2025-10-02
Update Adamant Tool Versions

dinkelk:
Update Alire and our Ada dependencies as well as redo and libadalang tools.


dinkelk:
All changes here were addressed in https://github.com/lasp/adamant/pull/121 instead.

PR #166, opened 2026-04-10
Avoid slow redo probes for nonexistant completion prefixes

dinkelk:
## What this fixes
This fixes slow/no-match bash completion for slash-separated redo target prefixes such as:

- `redo build/src/component-y<TAB><TAB>`
- `redo doc/build/pdf/command_router_nope<TAB><TAB>`

Before this change, those completions could feel hung because the completion script recursed into virtual target prefixes and then probed synthetic redo paths like `build/what_predefined` and `build/src/what_predefined`, each taking the slow failure path.

## Root cause
`env/redo_completion.sh` was treating path-shaped prefixes from cached `redo what` output as if they were real redo directories.

That is correct for real filesystem directories, but wrong for virtual target paths such as:

- `build/...`
- `doc/build/...`

## Fix
When completion already has cached `redo what` output for a real directory, it now treats slash-separated target prefixes as a virtual target namespace and narrows that cached target list by prefix instead of probing deeper synthetic redo paths.

## Test Coverage
Updated `redo/test/persistent_cache/test_persistent_cache.sh` so the main suite now:
- uses `command_router`, which exists in this tree (does not require adamant_example)
- checks current cache-refresh behavior for new `.do` files
- includes the new virtual-prefix completion regression coverage

### DIFF
--- a/env/redo_completion.sh
+++ b/env/redo_completion.sh
@@ -160,6 +160,36 @@ $(compgen -d "$full_arg" | sed 's/$/\//')" | grep -v '^$')
         try_trim_leading_dir "$path" "$arg"
     }
 
+    # When we already have a cached target list for a real redo directory,
+    # treat slash-separated target prefixes as a virtual namespace and keep
+    # narrowing the cached targets instead of probing synthetic subpaths like
+    # build/what_predefined (which can be very slow to fail).
+    try_virtual_recurse () {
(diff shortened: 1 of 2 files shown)


Jbsco (approved):
Clean, slow path optimizations look good, `cached_targets` seems to save on re-reads - Nice test additions, especially 34!

PR #168, opened 2026-04-15
Fix off-by-one in Get_Field for nested packed type fields

dinkelk:
## Summary                                                                                                                                                               
                                                            
  Fixes off-by-one errors in the generated `Get_Field` function for both packed record and packed array validation templates. In both cases, `Get_Field` and `Valid` disagreed on field numbering, causing `Get_Field` to read the wrong field (or out-of-bounds memory) when called with the errant field number returned by `Valid`.
                                                                                                                                                                           
  **Record template** (`gen/templates/record/name-validation.adb`): The generated code used `Field - (start_field_number + 1)` instead of `Field - (start_field_number -   1)`. For a field starting at position 17, this produced `Field - 18` instead of `Field - 16`, underflowing in `Unsigned_32` arithmetic. The bug is silent due to `pragma Suppress (Range_Check)` in the generated code, which is needed because the program reads whatever memory is at the wrapped address which is known to be out of range.                                                             
                                                                      
  **Array template** (`gen/templates/array/name-validation.adb`): The element index used `Field / num_fields` instead of `(Field - 1) / num_fields`. This selected the wrong array element when the field number was an exact multiple of `num_fields` (e.g., field 3 in a 3-sub-field element mapped to element 1 instead of element 0).
                                                                                                                                                                           
  Both bugs stem from the same root cause: sign/precedence errors when inverting the `Valid` function's field numbering formula.

### DIFF
--- a/gen/templates/array/name-validation.adb
+++ b/gen/templates/array/name-validation.adb
@@ -227,7 +227,7 @@ package body {{ name }}.Validation is
 {% if endianness in ["either", "big"] %}
    function Get_Field (Src : in T; Field : in Interfaces.Unsigned_32) return Basic_Types.Poly_Type is
 {% if element.is_packed_type %}
-      Idx : constant Constrained_Index_Type := Constrained_Index_Type'First + Unconstrained_Index_Type (Field / {{ element.type_model.num_fields }});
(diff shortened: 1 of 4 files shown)


Jbsco (approved):
Really great find! The coincidental pass for byte arrays (`[others => 0]`) against zeroed memory is an interesting failure mask.
Fix looks mathematically sound, templates appear consistent, and test hardening is appropriate.

PR #170, opened 2026-04-17
Enforce big-endian packed types for ID'd entities

dinkelk:
Adamant does not currently allow little-endian packed types for ID'd entities, including commands, parameters, etc.  However, there is no enforcement of this except for compilation error of incorrectly generated code. This PR provides a formal enforcement of big-endian-only ID'd entities. This ensures that code can not be generated for little-endian ID'd entities. If support for little-endian is added later, this pattern allows us to do so formally and incrementally.

## Summary

- Add `is_little_endian` property to the `datatype` class in `variable.py`, set in each packed-type branch (`T_Le`, `Volatile_T_Le`, `Atomic_T_Le`, `Register_T_Le`). For non-packed types it remains `False`.
- Expose through `ided_entity` so templates and model code can query endianness cleanly.
- Enforce big-endian for all ided entities by raising `ModelException` when a little-endian type is used. All code generation templates for commands, events, parameters, faults, data products, packets, and data dependencies hardcode big-endian `Serialization`, `Validation`, and `Representation` — using an LE type would produce silently wrong code.

The exception to this is the CCSDS Product Extractor component, which allows little-endian data products. Its templating already handles this case properly and this feature has been carried forward.

### DIFF
--- a/gen/models/submodels/ided_suite.py
+++ b/gen/models/submodels/ided_suite.py
@@ -21,6 +21,7 @@ class ided_entity(renderable_object):
         default_value=None,
         variable_types_allowed=False,
         packed_types_required=False,
+        little_endian_allowed=False,
         suite=None,
     ):
         self.name = ada.formatType(name)
@@ -68,6 +69,19 @@ class ided_entity(renderable_object):
                 + "'. Variable length types are not (yet) supported. Consider using a static sized packed type instead."
             )
 
+        # Little-endian packed types are not supported for most ided
+        # entities. The code generation templates assume big-endian for
+        # serialization, validation, and representation of these types.
+        if type and self.is_little_endian and not little_endian_allowed:
+            raise ModelException(
+                "Entity: '"
+                + str(self.name)
+                + "' uses little-endian type '"
+                + str(self.type)
+                + "'. Little-endian types are not supported for this entity. "
+                + "Use the big-endian variant (.T) instead."
+            )
+
         # We don't allow volatile datatypes in ided entities.
         if type and self.datatype.is_volatile_type:
             raise ModelException(
@@ -111,6 +125,12 @@ class ided_entity(renderable_object):
     def has_packed_type(self):
         if self.datatype:
             return self.datatype.is_packed_type
+        return False
+
+    @property
+    def is_little_endian(self):
+        if self.datatype:
+            return self.datatype.is_little_endian
         return False
 
     @property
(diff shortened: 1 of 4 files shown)


gen/models/submodels/ided_suite.py
```diff
@@ -68,6 +69,19 @@ def __init__(
...
+        # entities. The code generation templates assume big-endian for
+        # serialization, validation, and representation of these types.
+        if type and self.is_little_endian and not little_endian_allowed:
+            raise ModelException(
+                "Entity: '"
+                + str(self.name)
+                + "' uses little-endian type '"
+                + str(self.type)
+                + "'. Little-endian types are not supported for this entity. "
+                + "Use the big-endian variant (e.g. .T instead of .T_Le) instead."
```
Jbsco: Minor: this reads a little odd with the doubled "instead", maybe something like

> "Use the big-endian variant (.T) instead."

gen/models/submodels/ided_suite.py
```diff
@@ -68,6 +69,19 @@ def __init__(
...
+        # Little-endian packed types are not supported for most ided
+        # entities. The code generation templates assume big-endian for
+        # serialization, validation, and representation of these types.
+        if type and self.is_little_endian and not little_endian_allowed:
+            raise ModelException(
+                "Entity: '"
+                + str(self.name)
+                + "' uses little-endian type '"
+                + str(self.type)
+                + "'. Little-endian types are not supported for this entity. "
```
Jbsco, line 81: Can `self.entity_name` be used for a more immediately informative/actionable error message?

> "Command 'X' uses little-endian type 'Y'..."

Could be more useful (if always available and set).

Jbsco (changes requested):

gen/models/submodels/ided_suite.py
dinkelk: entity_name is not part of self... but maybe we can do introspection on self to see if it is actually an inherited type?

Jbsco: Ah, it's set at line 229 in `_initialize`. Maybe not worth it just for a name if plumbing is needed.

dinkelk: Yeah, a bit of plumbing is needed. I don't think it is worth it. This is not an error we encounter every day, so the user will likely be aware where they went astray.

Jbsco (approved):

PR #171, opened 2026-04-20
Validate byte arrays instead of typed records (ARM 13.9.1(12))

dinkelk:
This PR changes packed record validation functions to use byte arrays as input instead of the packed type itself. This is a safer practice than passing the packed type, which could be range checked upon subprogram call (but isn't, in practice, with GNAT compilers). There is no functionality added or subtracted with this PR, just an altered pattern, which no longer relies on implementation-defined behavior for functionality.

## Summary of Changes

Addresses ARM 13.9.1(12) denotes that passing a packed record with invalid scalar representations (NaN floats, out-of-range enums) as `in T` is implementation-defined behavior. GNAT block-copies composites, which is safe in Adamant usage today, but a compiler that copies field-by-field could trigger erroneous execution before `Valid` even begins.

- `Validation.Valid` and `Get_Field` now accept `Serialization.Byte_Array` instead of typed records. Internal overlay with `Import, Alignment => 1` materializes the record without a parameter-passing copy.
- `Valid_Le` / `Get_Field_Le` naming for little-endian (byte-array subtypes are the same Ada type, so overloading doesn't work).
- `Valid(U)` and `Valid(Unconstrained)` removed — dead code with no external callers.
- Component template: both fixed-length and variable-length commands/parameters validate raw bytes **before** `From_Byte_Array`. Variable-length path includes a `pragma Compile_Time_Error` proving the max-sized rename fits in the command buffer.
- `ccsds_product_extractor` and `seq_runtime` updated to use byte-array validation.
- User guide example updated. Test expectations updated where the validate-before-deserialize reordering changed error classification.


cbrisebois-lasp (approved):
Looks good!

Jbsco (approved):

PR #172, opened 2026-04-21
Add ignored and unconnected connector counts to assembly HTML summaries

dinkelk:
## Summary

Add assembly-level counts for:

- unconnected connectors
- ignored connectors

and surface both in the summary blocks for:

- `name_components.html`
- `name_connections.html`

## Semantics

Counts are based on connector slots / indices:

- **ignored** = connector slot marked ignored
- **unconnected** = connector slot that is neither connected nor ignored


Jbsco (approved):
Looks like counts now increment under `shallow_load` as well. Seems like intended behavior, increments separated from that state makes sense.

PR #174, opened 2026-04-24
Eliminate set/hash-order nondeterminism in generators and build rules

dinkelk:
## Summary

Audits `gen/`, `redo/`, and per-component `gen/models/` for patterns whose iteration order depends on Python's `PYTHONHASHSEED`, and replaces them with insertion-order-preserving equivalents. Each one is a latent reproducibility hazard: outputs that transitively depend on a set's iteration order can differ across Python processes even though inputs are identical.

## Changes

- `list(set(X))` -> `list(dict.fromkeys(X))` throughout. Same deduplication semantics, but `dict` preserves insertion order so the result is stable across processes.
- `gen/models/view.py`: rewrite two set-operation-into-dict-insertion patterns in `assembly_intersection`/`assembly_union` to iterate ordered dict keys filtered by set membership, instead of iterating the set result directly. These were feeding `data_dependencies` dict insertion order, which flowed into generated Ada code ordering.
- `gen/models/component.py`: same fix as `gen/models/assembly.py` for the component-level submodel loader (`for f in list(set(submodel_files))`), which had the same structural issue as the assembly submodel loader but for component submodels. Also cleans up `complex_type_models` to do dedup without going through a set.
- `gen/models/base.py` `unique()` helper: same swap.

Membership-only uses of `set(...)` (sets that are only queried with `in`, never iterated) are left alone -- those are already order-insensitive.

## Behavioral impact

No change to the element set of any affected container. No change to any logic that uses the containers for `in` membership. Templates that iterate `data_dependencies` built by `assembly_intersection`/`assembly_union` may produce entries in a different but now stable order -- blocks of `data_dependencies`-driven entries in generated Ada will be permuted relative to a prior nondeterministic run, not added or removed.

### DIFF
--- a/gen/models/view.py
+++ b/gen/models/view.py
@@ -65,7 +65,7 @@ def check_connector_names(filter_obj, assm):
     # Include non-indexed names in the list too. ie. if Tick_Divider_Instance.Tick_T_Send[3]
     # is in the list also include Tick_Divider_Instance.Tick_T_Send
     connector_names_no_indexes = [re.sub(r'\[[^\]]*\]$', '', x) for x in connector_names if x.endswith(']')]
-    connector_names = list(set(connector_names + connector_names_no_indexes))
+    connector_names = list(dict.fromkeys(connector_names + connector_names_no_indexes))
     for t in filter_obj.item_list:
         if t not in connector_names:
             warn(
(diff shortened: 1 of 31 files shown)


gen/models/view.py
```diff
@@ -614,13 +614,13 @@ def prune(assm_to_filter):
     assm_to_filter.data_dependencies = new_data_dependencies
     assm_to_filter.num_data_dependencies = sum(len(dd_list) for dd_list in new_data_dependencies.values())
 
-    # Cleanup 3: Remove components for which there is no connectors or data dependencies (if selected):
+    # Cleanup 3: Remove components for which there is no connectors or data dependencies (if selected).
```
cbrisebois-lasp: Line ends with . while other # Cleanup ... comments end with a :, turbo nit, but an inconsistency none-the-less.

cbrisebois-lasp (commented):
There is some commenting inconsistency here. My agents have been consistently ignoring commenting guides/skills so this feels eerily familiar.

Jbsco (approved):

Issue #16, opened 2024-01-22
Embedded Unit Testing


dinkelk:
New Feature

Add support for running unit test framework (even in a limited way) on embedded targets.

dinkelk:
@@MyApaulogies is working on this.

dinkelk:
https://blog.adacore.com/enhancing-ada-embedded-development-the-power-of-real-time-logging-with-rtt

dinkelk:
Capability added in https://github.com/lasp/adamant/pull/175.

PR #175, opened 2026-04-28
Make unit-test framework cross-compile-ready

dinkelk:
## Summary

Addresses: https://github.com/lasp/adamant/issues/16

Framework-level additions that enable Adamant component unit tests to be compiled against bareboard runtimes without dragging in host runtime services that don't exist there. Scope is target-neutral framework support — Linux tests still pass unchanged. The project-specific build glue (Jorvik configuration pragma, cross AUnit gpr, renode launcher, etc.) lives in the consuming project, not here.

### Split target-specific unit_test support packages

`file_logger`, `smart_assert`, `termination_handler`, and a new `test_clock` helper now have `Linux/` and `bb/` body subdirs picked by the build system's path filter. Public API is unchanged; only internal implementations differ per target.

  - `file_logger`: Linux body keeps a per-`Instance` `Ada.Text_IO.File_Type` (multi-logger safe); bareboard body routes `Log` to single-arg `Put_Line` which the embedded runtime maps to UART.
  - `smart_assert`: assertion failure messages route through a new target-split `Smart_Assert_Error_Io.Put_Line`. Linux uses `Standard_Error` (so `redo` doesn't swallow them); bareboard uses single-arg `Put_Line` (the stripped embedded `Ada.Text_IO` has no `File_Type` or `Standard_Error`).
  - `termination_handler`: Linux body installs a fallback handler via `Ada.Task_Termination.Set_Dependents_Fallback_Handler` and prints abnormal/exception terminations to `Standard_Error`; bareboard body is a no-op (the embedded runtime does not expose `Cause_Of_Termination`).
  - `test_clock`: new helper. Linux returns an `Image (Clock)` -based string in `Init_Logging` log lines; bareboard returns `""` (the embedded runtime has `Ada.Calendar` but not `Ada.Calendar.Formatting`).

The test scaffold template (`gen/templates/tests/name.adb`) is also updated to use these splits and the new `Test_Clock` helper.

### `Tester_Allocator` generic

Replaces the unconditional `new Component...Tester.Instance` / `Safe_Deallocator.Deallocate_If_Testing` pair in the test scaffold with a target-split allocator. Linux body heap-allocates each scenario fresh (matches historical behavior; no state leaks across scenarios). Bareboard body returns `'Access` of an instantiation-local aliased static instance — required because Jorvik forbids both heap allocation of types containing protected components (`No_Protected_Type_Allocators`) and function-local protected objects (`No_Local_Protected_Objects`). The user's `Set_Up_Test` re-runs `Init_Base` / `Init` / `Connect` per scenario so the static-storage reuse is correct.

### `redo test` cross-aware

`build_test.py` now consults the target's `is_cross()` predicate (new on `arm_bare_board` and `riscv_bare_board` target classes via a small base-class addition) and refuses to host-execute the resulting ELF when the target is cross.

### Init / Destroy semantics on Fifo, History, Circular_Buffer

Each data structure's `Destroy` resets internal state so the `Instance` "looks freshly Init'd" and the next `Init` can short-circuit the heap allocation when memory is already bound. On Linux `Safe_Deallocator` actually frees and the next `Init` reallocates; on bareboard the deallocator is a no-op so the buffer stays bound and `Init` skips the `new`. This matches the existing static-Tester reuse pattern that test fixtures rely on.

`Circular_Buffer.Queue_Base` now has its own `Destroy` override so `Item_Count` / `Item_Max_Count` are reset alongside the base `Head` / `Count` / `Max_Count`.

## Related

- xmera-components: https://github.com/lasp/adamant-xmera-components/pull/22


redo/targets/arm_bare_board.py
```diff
@@ -6,8 +6,11 @@
 
 class arm_bare_board(gprbuild_target_base):
     def path_files(self):
-        """ARM bare board targets contain the 32bit and bb (bareboard) path files."""
-        return super(arm_bare_board, self).path_files() + ["32bit", "bb"]
+        """ARM bare board targets contain the 32bit and bb (bareboard) path files.
+        bb_production marks bareboard sources whose semantics differ between production
+        and cross unit-test contexts (e.g. Safe_Deallocator). Cross test targets must
+        override path_files to drop bb_production and add bb_test."""
```
Jbsco, line 12: Minor and non-blocking, but might be worth providing a shippable example of an overriding target class in the docstring:
```python
def path_files(self):
    """[...] and add bb_test, e.g.:

        class arm_bare_board_test(arm_bare_board):
            def path_files(self):
                return [p for p in super().path_files() if p != "bb_production"] + ["bb_test"]
    """
    return super(arm_bare_board, self).path_files() + ["32bit", "bb", "bb_production"]
```
Could also include a similar example for `redo/targets/riscv_bare_board.py:8-11`.

Jbsco (changes requested):
Overall separation of host vs bareboard appears good, and well commented!
Only minor suggestion for comments in targets python and one more general question:
Would a link-only cross-compile smoke test under gen/test/cross_smoke/ (TARGET=arm_bare_board_test, stub component that touches the new bb/ bodies + Destroy resets, asserts .elf produces) be worth adding? Could catch future regressions in the cross paths in Adamant's own CI rather than relying on downstream consumers.

Jbsco (approved):

dinkelk:
@Jbsco - We are currently using https://github.com/lasp/adamant_example for this purpose and I think we should stick with that for now instead of having to choose a cross target to ship with Adamant. However, we should expand cross unit tests to Adamant Example repo in the near future!

PR #176, opened 2026-05-01
Fix packed array validation template alignment issue

dinkelk:
## Summary

The packed array Validation autocode (`Valid` / `Valid_Le` / `Get_Field` / `Get_Field_Le`, generated from `gen/templates/array/name-validation.adb`) overlaid the typed view directly on the input `Bytes`:

```ada
R : T with Import, Convention => Ada, Address => Bytes'Address;
```

If `Bytes`' runtime address didn't satisfy `T'Alignment`, this is **erroneous** per Ada 2022 RM 13.3(13/3). In practice that means silent garbage / right-answer-by-luck on x86_64 Linux, and traps / garbage-bit reads on alignment-strict bareboard targets like RV32IMAF.

Note that this issue does not affect packed records, since packed records all have Alignment of 1. GNAT does not allow compilation of packed arrays with an Alignment set to 1, which is the important difference.

## What changed

`Valid` / `Get_Field` now picks one of three runtime patterns based on the element shape (computed in `gen/models/array.py` as `required_alignment`):

- **Packed-record element** (`T'Alignment = 1`) — direct overlay, no copy needed.
- **Byte-aligned primitive** (`Short_Float`, `Unsigned_32`, `Long_Float`, …) — fast path overlays directly when `Bytes`' runtime address already satisfies `T'Alignment`; slow path copies into an aligned local `Byte_Array` and overlays on the copy.
- **Sub-byte primitive** (e.g. `T'Component_Size = 10`) — always copy into the aligned local. GNAT's bit-extract codegen still issues word loads under the hood, so a misaligned direct overlay reads garbage on bareboard.

`required_alignment` is computed in the array model and asserted to be a power of 2 at codegen time. Combined with the GNAT-RM 10.1 guarantee that "all default alignments are always a power of 2," a single `pragma Compile_Time_Error (T'Alignment > required_alignment, …)` at every overlay site is sufficient as a safety check (pow2-and-pow2-and-greater-or-equal implies "our literal is a multiple of `T'Alignment`," the second half of the RM 13.3(13/3) safety condition).

The record validation template gets a parallel `pragma Compile_Time_Error (T'Alignment /= 1, …)` at every overlay site to lock in the bit-packed-record codegen invariant.

## New tests

- `gen/test/arrays/test/test.adb` — added a "Validation misalignment tests" section that overlays each test array (Float_Array, Simple_Array, Unaligned_Array, Complex_Float_Array, Complex_Array_Le) at byte offsets 0..3 and exercises `Valid` / `Valid_Le` and `Get_Field` / `Get_Field_Le`.
- `gen/test/arrays/test_misalign/` — standalone bareboard variant of the same coverage, sized to fit the embedded runtime's secondary stack. Reports in AUnit text-reporter format so cross-test runners can scrape pass/fail.
- `gen/test/records/test_misalign/` — companion that exercises the record template's direct overlay (Always_Valid_Simple, Aa, Ff with Short_Float + Long_Float fields) at byte offsets 0..3, BE + LE.

### DIFF
--- a/gen/models/array.py
+++ b/gen/models/array.py
@@ -108,6 +108,71 @@ class array(type):
                 + "total packed array itself must be byte aligned."
             )
 
+        # Compute a static literal that is an upper bound on the
+        # alignment GNAT will pick for this packed array's T (and
+        # T_Le) types. The validation autocode uses this literal in
+        # `with Alignment => ...` aspects on local Byte_Array copies
+        # that are then overlaid by `R : T`. The safety condition for
+        # the overlay (RM 13.3(13/3)) is met when `our_literal >=
+        # T'Alignment` and `our_literal mod T'Alignment = 0`. Both
+        # follow from "our literal is a power of 2 >= T'Alignment"
+        # because GNAT-RM 10.1 guarantees T'Alignment is also a power
+        # of 2 (any pow2 >= a smaller pow2 is a multiple of it).
+        #
+        # Records always have Alignment => 1 by GNAT-RM 13.4. For
+        # everything else we cannot match both targets exactly with
+        # one formula, so we pick a literal that is an upper
+        # bound on both. Specifically:
+        #   * byte-aligned: bytes_per_elem & -bytes_per_elem
+        #   * sub-byte: next_pow2 of total array byte size
+        # Over-aligning is harmless beyond a small stack-space cost
+        # for the local Byte_Array copy. Under-aligning would be
+        # erroneous. The pragma at every overlay site catches any
+        # under-alignment at compile time.
+        #
+        # The `n & -n` idiom isolates the lowest set bit, which for
+        # any positive integer is the greatest power of 2 dividing
+        # it.
+        if self.element.is_packed_type and not isinstance(
+            self.element.type_model, array
+        ):
+            # Packed-record element: T'Alignment = 1, also enforced
+            # by the parallel pragma Compile_Time_Error in the
+            # record-validation template.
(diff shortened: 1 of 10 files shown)


gen/test/arrays/test_misalign/test.adb
```diff
@@ -0,0 +1,342 @@
+--------------------------------------------------------------------------------
+-- Misalignment-resilience test for the packed-arrayValidation autocode,
```
Jbsco: Minor nit, missing space:
"packed-arrayValidation" -> "packed-array Validation"

Jbsco (changes requested):

Jbsco (approved):

gen/models/array.py
```diff
@@ -108,6 +108,53 @@ def _load(self):
...
+        #   - sub-byte primitive:     next_pow2(ceil(total_bits/8))
+        # Note that we cannot check if our literal is
+        # exactly equal to T'Alignment, since alignment differs on
+        # different architectures. Instead we check for compatibility.
+        # The validation template emits a `pragma Compile_Time_Error`
+        # at every overlay site that fails the build if our static
+        # literal is < T'Alignment or not a multiple of T'Alignment,
+        # so any future GNAT change to this rule fails loudly at
+        # compile time.
+        if self.element.is_packed_type:
```
dinkelk: This is true only if the packed type is a record.

dinkelk: If it is a packed array it needs to use the alignment of the component type

PR #177, opened 2026-05-04
Fix .C array Pack/Unpack byte-swap across SSO boundary

dinkelk:
## Summary

- Fixes a value-dependent CONSTRAINT_ERROR in autocoder-generated `.C.Unpack` for arrays of `Short_Float` (and any non-packed-element array whose component crosses the BE/LE storage boundary).
- Routes the four cross-SSO bodies in `gen/templates/array/name-c.adb` through the parent package's `Pack`/`Unpack` (which use array-type-conversion). GNAT honors `Scalar_Storage_Order` for type conversions; it does not honor it for the iterated component association `[for J in Src'Range => Src(J)]` on every target (observed broken on `riscv32-elf`).
- Adds a host-side round-trip pin in `gen/test/arrays/test/test.adb` for `Float_Array.C.Pack`/`Unpack`.

## Background

This issue was experienced while executing `Packed_F32x3.C.Unpack` while reading a perfectly valid `9.45784093e-4` (= bit pattern `0x3A77EE7F`). Its byte-reverse `0x7FEE773A` is a NaN, and the runtime's `'Valid` check was firing on the unswapped bytes — strong evidence that the iterated-aggregate path was not byte-swapping the reads as the type's `Scalar_Storage_Order => System.High_Order_First` requires. The Pack direction emits the swap correctly on the same target, so producers wrote correctly-BE-encoded bytes that the consumer then misread.

### DIFF
--- a/gen/templates/array/name-c.adb
+++ b/gen/templates/array/name-c.adb
@@ -30,7 +30,7 @@ package body {{ name }}.C is
 {% if element.is_packed_type %}
       return [for J in Src'Range => {{ element.type_package }}.C.Pack (Src (J))];
 {% else %}
-      return [for J in Src'Range => Src (J)];
+      return {{ name }}.Pack (To_Ada (Src));
 {% endif %}
    end Pack;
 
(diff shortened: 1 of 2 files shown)


Jbsco (approved):

cbrisebois-lasp (approved):
Good catch! Spooky though; these target specific inconsistencies are starting to get very tiresome

PR #180, opened 2026-06-01
Add Parameter_Table_Forwarder component and enhance the parameter table subsystem

dinkelk:
## Summary

Adds a new generic **`Parameter_Table_Forwarder`** component and, in support of it, overhauls the shared parameter-table machinery (`Parameters`, `Parameter_Store`, the `Ccsds_Parameter_Table_Router` model tooling, and supporting types). The headline feature is the forwarder; the rest of the change generalizes the parameter subsystem so that components owning large, custom-layout tables are first-class citizens alongside the existing per-parameter pipeline.

## Headline: `Parameter_Table_Forwarder`

A validating proxy that sits between an upstream parameter-table source (`Ccsds_Parameter_Table_Router` or an upstream `Parameter_Store`) and a single downstream component that owns its own packed-record table layout.

Mission components needing big custom tables can't use the per-parameter `Parameter_Update.T` pipeline (individual parameters are bounded by the parameter buffer size). Before this, such components would each have to reimplement length checks, CRC validation, dump commands, packets, and status data products. The forwarder absorbs that boilerplate:

- **Active**, owns no table bytes — pure middleware. The downstream component is the authoritative table owner.
- Forwards all four operations (`Set`, `Validate`, `Get_Copy`, `Get_Pointer`) to the downstream via a single `recv_sync` handler + release send. Synchronous forwarding — no "waiting on downstream ack" state machine.
- Standard boilerplate handled for free: length + CRC validation, `Dump_Parameter_Table` command, `Dump_Parameters_On_Change` auto-dump, `Table_Status` data product, and a full event set (length/CRC rejections, table updated/validated/fetched, downstream rejections, dump failures, drops).
- **Large-table dump path:** dumps via `Memory_Dump_Send` → `Memory_Packetizer`, which chunks tables exceeding `Packet_Buffer_Size` into multiple `Packet.T`s sharing one APID. No single-`Packet.T` size limit.
- Minimal downstream contract: one `Parameters_Memory_Region` `recv_sync` handler that dispatches on operation, plus one release send. No CRC, length checks, events, commands, packets, or status DPs to reimplement.

## Parameter-subsystem overhaul

- **`Parameter_Table_Util` shared helper** (`src/types/parameter/`): extracts the duplicated table-CRC computation and memory-region header slicing out of `Parameters` and `Parameter_Store` into one place; new consumers (the forwarder) reuse it.
- **`Get_Pointer` operation** added to `Parameter_Table_Operation_Type`, and the legacy `Get` renamed to `Get_Copy` at every call site:
  - `Get_Copy`: existing semantics — copy current table bytes into the caller's region.
  - `Get_Pointer`: zero-copy alternative for large tables — return a region pointing at the component's own buffer (read-valid until the next `Set`). `Parameters` rejects it (its bytes are scattered across per-parameter records); `Parameter_Store` and the forwarder support it.
- **`Memory_Dump` send pathway in `Parameter_Store`**: large stored tables that exceed `Packet.T` can now dump through a downstream `Memory_Packetizer` instead of a single `Packet.T`. The `Packet.T` path remains the default; exactly one of the two is wired per instance.
- **`Byte_Array_Pointer.Copy_From`** (`src/types/memory/`): a procedural counterpart to `Copy_To` that copies bytes out of a pointer in place, avoiding the unconstrained-`Byte_Array` secondary-stack temporary that `To_Byte_Array` returns — which overflows the secondary stack for large (~10 KB) tables. `Parameter_Store`'s `Set` path now uses it.
- **Per-component-type dump-packet type resolution** (`parameters_packets` model): replaces the original lookup with a per-component-type dispatch (`Parameters` self-init, `Parameter_Store` via `Parameter_Manager`, `Parameter_Store` via router named `table_id`). Each strategy asserts its component-type contract; unresolved packets emit a soft warning and stay typeless rather than mis-binding. Packets dumped via `Memory_Dump_Send` are deliberately left typeless (they downlink as chunked byte blocks, mirroring `Memory_Dumper`'s `Memory_Dump_Packet`), since a flat-record type is fictional there.
- **Python generator for `parameter_table_router_table`**: emits a `<assembly>_parameter_table_router_table.py` alongside the existing `.ads`, exposing each router `table_id` and its destination instances so ground tooling can resolve which numeric `table_id` targets a given downstream component without hand-mirroring the YAML.
- **`Generic_Staged_Variable`** added to `Protected_Variables` (+ unit tests): a reusable protected-object pattern for handing a value from a writer task to an owner task (`Stage` / `Is_Staged` / `Copy_From_Staged`), used by downstream consumers that stage a table on one task and apply it on another.

## Testing

- `Parameter_Table_Forwarder` unit tests: Set/Validate/Get success and error paths, downstream rejections, the dump command (success / failure / no-connector), `Dump_Parameters_On_Change`, queue-full drops, invalid command, and initial `Table_Status` — with full event and data-product history verification on every path.
- `Protected_Variables` generics unit tests covering all five generics including `Generic_Staged_Variable`.
- Migrations of `Parameters` / `Parameter_Store` to the shared helper are behavior-preserving; existing small-table (`Packet.T`) dump behavior is unchanged. The large-table (`Memory_Dump`) path is purely additive.


src/components/parameter_table_forwarder/parameter_table_forwarder.parameters_packets.yaml
```diff
@@ -0,0 +1,8 @@
+---
+description: Packets for the Parameter Table Forwarder component.
+packets:
+  - name: Active_Parameters
+    description: |
+      Snapshot of the current parameter table fetched from the downstream
+      component at the time of dump, prefixed with a freshly computed CRC for
```
Jbsco: As far as I can tell, `Send_Active_Parameters_Packet` emits `Stored_Header` (the CRC stored at the last Set), not a freshly recomputed CRC. This description should probably be updated from "freshly computed CRC" to "stored table CRC", unless I'm misunderstanding.

src/components/parameter_table_forwarder/test/component-parameter_table_forwarder-implementation-tester.ads
```diff
@@ -0,0 +1,231 @@
...
+   -- A parameter memory region was dropped due to a full queue.
+   overriding procedure Memory_Region_Dropped (Self : in out Instance; Arg : in Parameters_Memory_Region.T);
+
+   -----------------------------------------------
+   -- Data product handler primitives:
+   -----------------------------------------------
+   -- Description:
+   --    Data products for the Parameter Table Forwarder component.
+   -- Version/CRC/timestamp/last-operation status of the active table as known
+   -- to the forwarder. Updated on every Set/Validate/Get/Dump operation.
```
Jbsco: Minor nit: Is this updated on Get/Dump? [component-parameter_table_forwarder-implementation.adb:321](https://github.com/lasp/adamant/pull/180/changes#diff-c91409fe1623e1b21a8ac6e1302e01a98b7b34948f7d28c02f1d7100ae0d397aR321) suggests Set/Validate only.

src/components/parameter_table_forwarder/test/component-parameter_table_forwarder-implementation-tester.adb
```diff
@@ -0,0 +1,392 @@
...
+      Self.Memory_Region_Dropped_History.Push (Arg);
+   end Memory_Region_Dropped;
+
+   -----------------------------------------------
+   -- Data product handler primitive:
+   -----------------------------------------------
+   -- Description:
+   --    Data products for the Parameter Table Forwarder component.
+   -- Version/CRC/timestamp/last-operation status of the active table as known
+   -- to the forwarder. Updated on every Set/Validate/Get/Dump operation.
```
Jbsco: Same here, comment seems to diverge from Set/Validate at [component-parameter_table_forwarder-implementation.adb:321](https://github.com/lasp/adamant/pull/180/changes#diff-c91409fe1623e1b21a8ac6e1302e01a98b7b34948f7d28c02f1d7100ae0d397aR321).

Jbsco (approved):
Overall approved with minor questions/nits on comment wording.

PR #185, opened 2026-07-20
Remove the GNAT-specific Value_Size aspect from templates and sources

cbrisebois-lasp:
Value_Size is a GNAT-only alias of Size and breaks Adamant's compiler portability; Size on first subtypes and the standard Object_Size cover every existing use, so the aspect adds nothing.

### DIFF
--- a/gen/templates/component/component-name.ads
+++ b/gen/templates/component/component-name.ads
@@ -351,9 +351,9 @@ private
       Index : Connector_Index_Type;
    end record
       with Size => 24,
-             Object_Size => 24,
-             Alignment => 1,
-             Volatile => False;
+           Object_Size => 24,
+           Alignment => 1,
+           Volatile => False;
 
    -- Define the size of the record to be compact to save space on
    -- the queue (24 bits).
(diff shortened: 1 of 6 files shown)


gen/templates/component/component-name.ads
```diff
@@ -352,7 +352,6 @@ private
    end record
       with Size => 24,
              Object_Size => 24,
```
dinkelk: can we fix the spacing here while we are at it?

dinkelk:
In modern Ada `Size`/`Object_Size` pair fully specify both the size a type takes to represent a value and the size a type takes in memory, respectively. Both aspects are in the Ada standard (`Size` has been standard Ada all along, and `Object_Size` started life as a GNAT extension but was standardized in Ada 2022, AI12-0059). `Size` is fully redundant with `Value_Size` and `Size` is preferred, since it is not GNAT-only, like `Value_Size` is. This change makes the usage clear within Adamant.

dinkelk (changes requested):

gen/templates/component/component-name.ads
cbrisebois-lasp: on it

dinkelk (approved):

PR #186, opened 2026-07-20
Bugfix/ccsds serial interface length wrap

cbrisebois-lasp:
### Fix modular wrap-around in the Ccsds_Serial_Interface packet length guard

**Summary**

A received CCSDS header whose Packet_Length field is 16#FFFF# crashes the Listener task with an unhandled Constraint_Error instead of being rejected with a Packet_Recv_Failed event. Because the listener task fails loudly and terminates, packet reception on the serial interface is lost until restart.

**Root cause**

In component-ccsds_serial_interface-implementation.adb, the listener validates the received length against the packet buffer before reading the packet data:

if (Packet.Header.Packet_Length + 1) > Packet.Data'Length then

Packet_Length is Interfaces.Unsigned_16, so the + 1 is computed with modular arithmetic. For the maximum value 16#FFFF# it wraps to 0, the guard passes, and the subsequent read indexes the packet buffer with the unguarded length:

Diagnostic_Uart.Get (Packet.Data (0 .. Natural (Packet.Header.Packet_Length)));

which raises Constraint_Error (range check) at this line. Every other oversized length (Data'Length .. 16#FFFE#) is caught by the guard and rejected cleanly; 16#FFFF# is the single value that defeats it. This is the first point at which the received length is inspected, so no earlier check can intercept it.

Note that GNAT SAS flagged this line previously; the finding was annotated as a false positive in 37e58e9 (the annotation's initialization rationale is correct, but the range hazard was real).

**Fix**

Compute the guard in Natural before the increment:

if (Natural (Packet.Header.Packet_Length) + 1) > Packet.Data'Length then

**Testing**

- Extended the test_read unit test: the input stream now appends a valid sync pattern followed by a header carrying Packet_Length => 16#FFFF# after the existing nominal packet. The test asserts the oversized packet is not forwarded and produces exactly one Packet_Recv_Failed event carrying the offending length.
- The test commit is ordered before the fix commit: at the first commit the suite fails with raised CONSTRAINT_ERROR : component-ccsds_serial_interface-implementation.adb:132 range check failed (listener task death), demonstrating the defect; with the fix applied the suite passes (1/1).
- redo style_all passes on the component and all test directories.


src/components/ccsds_serial_interface/component-ccsds_serial_interface-implementation.adb
```diff
@@ -125,7 +125,7 @@ package body Component.Ccsds_Serial_Interface.Implementation is
          -- First read the header:
          Diagnostic_Uart.Get (Header_Bytes);
 
-         if (Packet.Header.Packet_Length + 1) > Packet.Data'Length then
+         if (Natural (Packet.Header.Packet_Length) + 1) > Packet.Data'Length then
```
dinkelk, line 134: Comment explaining the type conversion here is good, so we avoid someone accidentally removing it again in the future.

src/components/ccsds_serial_interface/test_read/ccsds_serial_interface_tests-implementation.adb
```diff
@@ -74,6 +75,20 @@ package body Ccsds_Serial_Interface_Tests.Implementation is
 
       -- Make sure that the data size matches what we expect:
       Ccsds_Space_Packet_Assert.Eq (Packet, Expected);
+
+      -- The input stream additionally contains a sync pattern followed by a
+      -- CCSDS primary header whose Packet_Length field is 16#FFFF#. The
+      -- resulting packet cannot fit in the packet buffer, so the listener must
+      -- reject it with a Packet_Recv_Failed event and must not forward it. A
+      -- Packet_Length of 16#FFFF# in particular wraps to zero if the length
```
dinkelk: this comment is already stale... it should be worded for a forward audience that has no knowledge of the past

dinkelk (changes requested):

dinkelk (approved):

PR #188, opened 2026-07-24
Make redo clean instant: rename build dirs, reclaim in background

dinkelk:
## Problem

On slow virtualized filesystems (Docker Desktop bind mounts on macOS via VirtioFS), `redo clean` / `redo clean_all` can take **many minutes** on a bogged down or full-disk system. Deletion is one metadata round-trip per file across the host<->VM bridge, and build trees contain thousands of files.

## Change

`redo clean` now:
1. **Renames** each `build/` dir (a single O(1) metadata op) into a per-invocation session directory under `<repo_root>/.build_trash/`, mirroring its original path:
   ```
   src/components/command_router/build
     -> .build_trash/<pid>_<token>/src/components/command_router/build
   ```
2. **Detaches one background reclaimer** (`rm -rf` via `Popen(start_new_session=True)`, null stdio) over all trash and returns immediately.

The tree is logically clean the moment clean returns; disk space follows in the background.

Safety/robustness:
- **Orphan sweep**: every clean adopts all existing `.build_trash/` entries into its reclaim batch, so trash from a reclaimer killed mid-flight (container restart) never outlives the next clean.
- **Fallbacks**: cross-filesystem rename failure (`EXDEV`) or build dirs outside the repo root fall back to synchronous `rm -rf`; not inside a git repo falls back to the old behavior entirely.
- **`DISABLE_FAST_CLEAN=1`** forces the old synchronous delete (CI / throwaway containers / need-disk-now).
- `.build_trash/` starts with `.` so every build-system walk (`util/filesystem.py` and clean's own walk) already skips it; git-ignored; safe to delete manually at any time.
- Unique session names (pid + random token) make overlapping/repeated cleans collision-free; racing reclaimers degrade gracefully (`rm -f` ignores ENOENT).

### DIFF
--- a/redo/database/_setup.py
+++ b/redo/database/_setup.py
@@ -1,5 +1,4 @@
 import os.path
-import git
 import random
 from util import error
 from util import redo_arg
@@ -111,10 +110,8 @@ class _build_path(OrderedDict):
 
 def _get_git_root(path):
     """Get the root directory of the adamant repository."""
-    try:
-        git_repo = git.Repo(path, search_parent_directories=True)
-        git_root = git_repo.git.rev_parse("--show-toplevel")
-    except BaseException:
+    git_root = filesystem.get_git_root(path)
+    if git_root is None:
         error.error_abort(
             "No valid git repository was found containing the directory: " + str(path)
         )
--- a/redo/rules/build_clean.py
+++ b/redo/rules/build_clean.py
@@ -1,9 +1,31 @@
+import os
 import os.path
-from util import redo
-from util import debug
-from base_classes.build_rule_base import build_rule_base
+import secrets
 import subprocess
 import concurrent.futures
+from util import redo
+from util import debug
+from util import filesystem
+from base_classes.build_rule_base import build_rule_base
+
+# Name of the top-level trash directory used by the fast clean. Build
+# directories are renamed into here and reclaimed by a detached background
+# process. The leading "." ensures the build system's directory walks
+# (see util/filesystem.py) never descend into it. The entire directory is
+# garbage and is always safe to delete manually.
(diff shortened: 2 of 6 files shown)


redo/database/_setup.py
```diff
@@ -111,10 +110,8 @@ def path_filename(target):
 
 def _get_git_root(path):
     """Get the root directory of the adamant repository."""
-    try:
```
cbrisebois-lasp, line 114: Just out of curiosity, did you actually time this to see if it was faster? timeit module or whatever?

redo/rules/build_clean.py
```diff
@@ -1,9 +1,31 @@
...
+
+# Name of the top-level trash directory used by the fast clean. Build
+# directories are renamed into here and reclaimed by a detached background
+# process. The leading "." ensures the build system's directory walks
+# (see util/filesystem.py) never descend into it. The entire directory is
+# garbage and is always safe to delete manually.
+TRASH_DIR_NAME = ".build_trash"
+
+
+def _fast_clean_disabled():
```
cbrisebois-lasp, line 19: idiomatic python naming, nice

cbrisebois-lasp (approved):
Honestly never looked into clean. This is a good fix

redo/database/_setup.py
dinkelk: The _get_git_root? No - this is just a refactor.

The PR as a whole? Yes. On my system sometimes cleans would take 12 minutes (no joke) and this brought it down to seconds. On a non-full-ssd system the last test I ran went from ~43 s to ~1.5 s.

Jbsco (approved):

PR #190, opened 2026-07-29
Add Stopwatch.Reporter utility and refactor Rate_Group to use it

dinkelk:
## Summary

Adds `Stopwatch.Reporter`, a child package of `Stopwatch` that pairs a wall-clock and CPU-execution stopwatch for timing a section of code, accumulates recent-maximum and all-time-maximum values for both, and reports them as a `Task_Timing_Report.T` (the same type `Rate_Group` publishes, so measurements from any component can share a downlink packet).

The `Rate_Group` component previously hand-rolled exactly this logic in its tick handler; the second commit refactors it to use the new package (net -38 lines) with identical behavior.

## Design notes

- **External wall times**: `Start`/`Stop` overloads accept an externally supplied wall clock time, supporting backdating a measurement to an incoming tick's timestamp so queue latency is included (the Rate_Group pattern), or supplying both endpoints from converted `Sys_Time` stamps.
- **Max-updated indication**: `Accumulate` optionally reports whether either all-time maximum was updated, supporting Rate_Group's time-exceeded events.
- **Per-operation reporting**: `Report_Last` carries the most recent measurement alongside the all-time maximums, for components that publish a report per operation rather than periodically.
- `Instance` is a public record, matching the parent `Stopwatch` style.

## Validation

- New unit test suite (`src/util/stopwatch/test`, 4 tests) exercising start/stop with injected times, accumulation and max-updated indications, both report variants, and both resets: 4/4 pass.
- All 5 existing `Rate_Group` unit tests pass unmodified against the refactor.
- `redo style` passes on both directories.


src/components/rate_group/component-rate_group-implementation.ads
```diff
@@ -29,13 +29,11 @@ private
 
    -- The component class instance record:
    type Instance is new Rate_Group.Base_Instance with record
-      -- Time spans for keeping track of maximum
-      -- execution data:
+      -- Timer for keeping track of maximum execution data. The wall timer
+      -- measures the cycle time, starting at the time stamp of the incoming
+      -- tick:
       Issue_Time_Exceeded_Events : Boolean := False;
```
Jbsco: Minor: seems like `Timer` should be above `Issue_Time_Exceeded_Events` based on the comment read.

src/util/stopwatch/stopwatch-reporter.adb
```diff
@@ -0,0 +1,89 @@
...
+   begin
+      Ignore := To_Delta_Time (Max_Wall_Time, To_Return.Max.Wall_Time);
+      Ignore := To_Delta_Time (Max_Execution_Time, To_Return.Max.Execution_Time);
+      Ignore := To_Delta_Time (Recent_Wall_Time, To_Return.Recent_Max.Wall_Time);
+      Ignore := To_Delta_Time (Recent_Execution_Time, To_Return.Recent_Max.Execution_Time);
+      return To_Return;
+   end To_Report;
+
+   function Report (Self : in Instance) return Task_Timing_Report.T is
+      (To_Report (Self.Max_Wall_Time, Self.Max_Execution_Time, Self.Recent_Max_Wall_Time, Self.Recent_Max_Execution_Time));
```
Jbsco: Named associations here and :74 `Report_Last` would prevent transposition and match convention elsewhere.

src/util/stopwatch/stopwatch-reporter.ads
```diff
@@ -0,0 +1,59 @@
+with Task_Timing_Report;
+
+-- A paired wall-clock and CPU-execution stopwatch for timing a section of
+-- code. The reporter accumulates recent-maximum and all-time-maximum (high
+-- water mark) values for both timers, and produces reports of the
+-- accumulated values as a Task_Timing_Report.T, suitable for publishing as
+-- a data product.
+package Stopwatch.Reporter is
+
+   type Instance is tagged record
```
Jbsco, line 10: Should this be moved to private?

src/components/rate_group/component-rate_group-implementation.adb
```diff
@@ -58,48 +55,26 @@ package body Component.Rate_Group.Implementation is
       -- Pet the watchdog to indicate this subprogram is executing.
       Self.Pet_T_Send_If_Connected ((Count => Arg.Count));
 
-      -- Stop execution timer:
-      Sw_Cpu.Stop;
+      -- Stop the timers:
       Stop_Wall_Time := Self.Sys_Time_T_Get;
+      Self.Timer.Stop (Wall_Stop_Time => To_Time (Stop_Wall_Time));
```
Jbsco: The Sys_Time call on the line above is itself a part of the measurement, am I reading that correctly? Any concerns with this?

Jbsco (changes requested):
Some callouts, also wondering if `timing_report.record.yaml` should move to `src/types/task/`, alongside `task_timing_report.record.yaml`.
And should the `Max_Cycle_Time_Exceeded` value assertion in the Rate_Group suite be restored as part of this?

src/util/stopwatch/stopwatch-reporter.ads
dinkelk: I thought about this, but chose to make public to be uniform with the parent class. None of the contained state is dangerous to read or write, and it might actually be useful to be able to modify it in certain cases. I could go either way, but chose historical legacy in this case.

src/components/rate_group/component-rate_group-implementation.ads
dinkelk: Fixed.

src/util/stopwatch/stopwatch-reporter.adb
dinkelk: Done.

src/components/rate_group/component-rate_group-implementation.adb
dinkelk: You're reading it right, and it was a behavior change from the original (which stopped the CPU timer before the fetch). Fixed by splitting Stop into Stop_Cpu_Timer/Stop_Wall_Timer so the rate group stops the CPU timer, fetches the time, then stops the wall timer. The fetch is back out of the CPU measurement.

dinkelk:
Thanks for the review. All addressed and squashed back to 2 commits:

- Named associations in the To_Report calls: done.
- Timer field moved under its comment: done.
- Sys_Time fetch in the CPU measurement: good catch, fixed with split Stop_Cpu_Timer/Stop_Wall_Timer primitives (see inline reply).
- timing_report.record.yaml moved to src/types/task next to task_timing_report.record.yaml.
- Max_Cycle_Time_Exceeded value assertion restored -- the wall side is deterministic in test, so it works again. Only the execution time values remain count-only (real CPU clock).
- Instance stays public per the inline thread.

src/util/stopwatch/stopwatch-reporter.adb
```diff
@@ -0,0 +1,102 @@
...
+   procedure Stop_Cpu_Timer (Self : in out Instance) is
+   begin
+      Self.Cpu_Timer.Stop;
+   end Stop_Cpu_Timer;
+
+   procedure Stop_Wall_Timer (Self : in out Instance; Wall_Stop_Time : in Ada.Real_Time.Time) is
+   begin
+      Self.Wall_Timer.Stop_Time := Wall_Stop_Time;
+      Self.Last_Wall_Time := Self.Wall_Timer.Result;
+      Self.Last_Execution_Time := Self.Cpu_Timer.Result;
```
dinkelk: Why is this captured here and not in Stop_Cpu_Timer? should this be moved to accumulate?

dinkelk: Good point, moved the CPU result capture into Stop_Cpu_Timer so each half stores its own timer's result. Kept the captures out of Accumulate though: that would couple Report_Last to accumulation and make Accumulate read live timers instead of the last completed measurement.

Jbsco (approved):
`src/components/rate_group/types/` is now an empty directory, except for `.all_path`, could be removed in this PR, but not blocking. Otherwise looks good!

dinkelk:
> `src/components/rate_group/types/` is now an empty directory, except for `.all_path`, could be removed in this PR, but not blocking. Otherwise looks good!

addressed.

PR #193, opened 2026-07-30
Fix implicit parent package discovery for multi-line package declarations

### LINKED ISSUE #191: Component source discovery drops packages whose declaration spans lines (SPARK aspect form)
### Summary
Component source-directory discovery matches package declarations with a regex of the form
`package .* is\s*$`, which assumes `package ... is` sits on a single line. The SPARK-Mode aspect on
a spec pushes `is` onto the next line:

```ada
package Component.My_Component.Implementation
   with SPARK_Mode => On
is   -- "is" is now on its own line; the single-line regex does not match
```

When the regex misses the package, the component's source directory is never added to
`XSOURCE_DIRS`, and `gprbuild` then fails to find the unit.

### Symptom
- The component's directory is absent from `XSOURCE_DIRS`.
- `gprbuild` reports the unit is not found / not part of the project.

### Root cause
The discovery regex assumes the `package` keyword and its `is` are on the same line. The SPARK
aspect form (and any multi-line package declaration) breaks that assumption.

### Proposed fix
Make discovery tolerant of a multi-line package declaration - match `package <name> ... is`
across newlines (or normalize whitespace/newlines before matching). This lets the idiomatic SPARK
aspect form (`with SPARK_Mode => On` on a spec) work without a workaround.

### Current workaround (delete once fixed)
Use `pragma SPARK_Mode (On)` inside the package body instead of the aspect form on the spec (keeps
the header on one line); or, when the aspect is on a generated spec you do not control, add a shim
package in the assembly with an explicit `with Component.My_Component;`.

Jbsco:
## Summary

Source-dependency scanning extracts a package's implicit parent dependency by matching `package .* is\s*$` against newline/semicolon-split statement fragments. Any package declaration where the name and the `is` keyword are not on the same line defeats the match, so the parent include is silently dropped, the parent's directory never lands on the source path, and gprbuild fails to find the parent unit.

The common way to hit this is the SPARK aspect form, which conventionally puts `is` on its own line:

```ada
package Component.Foo.Implementation
   with SPARK_Mode => On
is
```

Symptom:

```
aspect_parent-child.ads:7:22: error: file "Aspect_Parent.ads" not found
```

Two adjacent latent cases fall out of the same match: `private package A.B.C is` never matched (leading keyword), and the single-line aspect form only produced a correct parent by accident of greedy stripping.

Closes #191.

## Fix

Extract the package name directly after the `package` keyword with a multiline regex over the file content, instead of requiring the trailing `is` on the same fragment:

```python
r"^\s*(?:private\s+)?package\s+(?:body\s+)?([A-Za-z][\w.]*)"
```

Parent derivation is unchanged. The now-unused `remove_postfix` helper is deleted.

## Validation

- Before/after harness over four declaration shapes (multiline aspect, single-line aspect, private child, plain body): old code loses the parent on the multiline-aspect and private-child cases; new code passes all four with no mangled names.
- End-to-end in the Adamant container: a child package in the aspect form with its parent in a separate directory fails to build before the change (`file "Aspect_Parent.ads" not found`) and builds cleanly after.
- That reproduction is added as `redo/test/aspect_compile/` (mirroring `spark_compile`'s structure, listed in the test README), so `redo` in that directory regresses this permanently: it only compiles when the implicit parent of a multi-line declaration is discovered.

Context: found while adding SPARK contracts to a component (the aspect form on a spec is the idiomatic SPARK entry point), where the failure presents as an inexplicable missing-source error two steps removed from its cause.

### DIFF
--- a/redo/test/README.md
+++ b/redo/test/README.md
@@ -8,4 +8,5 @@ This directory contains a set of unit tests which ensure the functionality of th
 
 The following is a description of what you can expect to find in the subdirectories of this directory.
 
+* `aspect_compile/` - a test which makes sure a child package declared with an aspect specification (which puts the `is` keyword on its own line) still has its implicit parent-package dependency discovered
 * `c_compile/` - a test which makes sure the compilation of c source code and linking with an ada main file is working properly
--- a/redo/test/README.md
+++ b/redo/test/README.md
@@ -10,3 +10,4 @@ The following is a description of what you can expect to find in the subdirector
 
 * `aspect_compile/` - a test which makes sure a child package declared with an aspect specification (which puts the `is` keyword on its own line) still has its implicit parent-package dependency discovered
 * `c_compile/` - a test which makes sure the compilation of c source code and linking with an ada main file is working properly
+* `source_dependencies/` - a unit test for the Ada source scanning in `redo/util/ada.py`, locking in the dependency extraction and body-dependency heuristics over a set of declaration shapes; runs under `redo test` like the python tests in `gen/test/`
--- a/redo/test/aspect_compile/aspect_parent-child.adb
+++ b/redo/test/aspect_compile/aspect_parent-child.adb
@@ -1,0 +1,7 @@
+package body Aspect_Parent.Child
+   with SPARK_Mode => On
+is
+
+   function Doubled_Base return Natural is (2 * Base_Value);
+
+end Aspect_Parent.Child;
--- a/redo/test/aspect_compile/aspect_parent-child.ads
+++ b/redo/test/aspect_compile/aspect_parent-child.ads
@@ -1,0 +1,13 @@
+-- Child package declared with an aspect specification. The aspect pushes the
+-- "is" keyword onto its own line, which is the shape that used to break the
+-- build system's implicit parent-package dependency extraction: the parent's
+-- directory never made it onto the source path and gprbuild failed to find
+-- Aspect_Parent. This test compiles only when that extraction handles a
+-- multi-line package declaration.
+package Aspect_Parent.Child
+   with SPARK_Mode => On
+is
+
+   function Doubled_Base return Natural;
+
+end Aspect_Parent.Child;
--- a/redo/test/aspect_compile/aspect_parent-private_child.ads
+++ b/redo/test/aspect_compile/aspect_parent-private_child.ads
@@ -1,0 +1,13 @@
+-- A private child package. The leading "private" keyword used to prevent the
+-- build system from recognizing this as a package declaration at all, so the
+-- implicit dependency on the parent was never recorded.
+--
+-- Deliberately body-less: every declaration in this directory must be one the
+-- old extraction could not parse. A single plain "package body X.Y is" here
+-- would recover the parent dependency on its own and mask the regression this
+-- directory exists to catch.
+private package Aspect_Parent.Private_Child is
+
+   Tripled_Base : constant Natural := 3 * Base_Value;
+
+end Aspect_Parent.Private_Child;
--- a/redo/test/aspect_compile/depends/aspect_parent.ads
+++ b/redo/test/aspect_compile/depends/aspect_parent.ads
@@ -1,0 +1,8 @@
+-- Parent package for the aspect_compile test. It lives in a separate
+-- directory so that it is only found if the build system correctly extracts
+-- the implicit parent dependency from the child package declaration below.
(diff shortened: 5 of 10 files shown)


dinkelk:
I ran the old and new extraction logic side by side over a set of declaration shapes to characterize exactly how the behavior changes. The fix looks correct, and the behavioral differences all fall in the new code's favor — beyond the multiline-aspect case this also fixes `private package A.B is` (never matched before due to the leading keyword), single-line aspects containing dots (old code emitted garbage parents like `'A.B with Abstract_State => (X.Y'`), mixed-case declarations (the old `remove_prefix`/`remove_postfix` were case-sensitive while the regex wasn't), and child-unit generic instantiations/renamings (`package A.B is new G (...);`, `package A.B renames C.D;`), whose parents are genuine implicit dependencies the old code missed.

While verifying, I found a handful of adjacent gaps and one small regression that would be good to fix (here or as follow-ups):

1. **Multi-line `with` clauses silently drop trailing units.** The `with`-statement parsing just above the fixed code uses the same newline/semicolon statement-split, so `with Foo,\n     Bar;` yields `['Foo', '']` — `Bar` is lost. This is the same failure class this PR fixes (silent missing dependency → missing source path → `file not found`), just in the other half of the function.

2. **`should_depend_on_adb` misses argument-less `Inline` aspects.** Its detection only matches `with Inline => True` with whitespace around `=>`. It misses `with Inline;` (legally equivalent to `=> True`) and `Inline=>True`. This one errs in the dangerous direction: a missed `.adb` dependency means stale objects under `-gnatn` rather than a loud build failure. It also only detects `generic` when alone on a line, missing a one-line `generic package ... is` header.

3. **Small regression: package declarations starting mid-line after a `;` no longer match.** The old statement-split on `[;/]` caught `pragma Elaborate_Body; package A.B is`; the new `^`-anchored regex misses it. No sane style does this, but it's cheap to recover with a `(?:^|;)\s*` prefix (non-capturing) if you want strict superset behavior.

4. **The dead safety filter stays dead.** `filter(lambda x: x.split() != 1, includes)` compares a list to an int, so it's always true and filters nothing. It was the intended last defense against garbage includes and has never worked — worth fixing to `len(x.split()) == 1` or deleting.

5. **Argument-less aspects on their own line become phantom includes.** `with Pure`, `with Ghost;`, `with Always_Terminates` (nothing triggers the `=>` filter) each produce a bogus include like `'Ghost'`. `source_db.try_get_sources` silently drops unresolvable names so it's usually invisible, but a real package by one of those names on the build path would create a spurious dependency. Given this PR's motivation is SPARK contracts, `with Ghost` will be common, so this will fire often.

6. **Test coverage:** the new `aspect_compile` test only locks the multiline-aspect case. A tiny unit test calling `get_source_dependencies` on a handful of fixture strings (private child, single-line aspect with dotted expressions, mixed case, instantiation/renaming) would lock in all the fixed behaviors for near-zero cost — cheaper than additional `redo`-level compile tests.

For completeness, the things I checked that are *not* problems: comments can't false-positive (`--` always precedes any commented `package`, and `^\s*` won't cross it); `\s*` spanning newlines under `re.MULTILINE` can't produce duplicate or wrong captures (`findall` is non-overlapping and results are deduped downstream); generic formal packages with `with` on a previous line match but have simple names, so no parent is derived; and this was the only `package ... is`-style matcher in the repo, so there's no missed twin elsewhere.

dinkelk (changes requested):
See previous comment

Jbsco:
- Statements now built by stripping comments per line, splitting on semicolons only, and collapsing to a single line. Matching those rather than raw text also keeps commented-out declarations out, where a `;` prefix on the raw text would not (points 1, 3, 5)
- `len(x.split()) == 1` applied (4)
- Statement helper: `generic` is `^\s*generic\b` so the one line header is caught, and inline is `\bwith\b.*\binline` - over-matching is a spurious body dep rather than a stale object from under-matching (2)
- aspect_compile test is body-less now. Agree on unit testing, but could follow the gen/test/*/test_py pattern, plain asserts with a test.do, which redo coverage_all already picks up

Jbsco:
Added the unit test as `redo/test/source_dependencies`, following the `test_py` pattern. Plain asserts over the declaration shapes above, fails against main's `ada.py`, and `redo test_all` discovers it so `coverage_all` runs it. Left `aspect_compile` as the end-to-end check.

dinkelk (approved):

PR #196, opened 2026-07-30
Add Product Store Component to enable save/restore for a selection of data products

owhite7128:
This component adds the functionality to save a set of data products to specified memory region. It has the ability to restore data products with different set times depending on the desired behavior.


src/components/product_store/product_store.component.yaml
```diff
@@ -0,0 +1,52 @@
+---
+description: |
+  The product store saves a predefined set of data products from the database to a byte array (memory region) provided at initialization, usually located in nonvolatile storage (i.e. MRAM). The set of data products to save is configured via an autocoded table provided at instantiation, produced from a stored_products.yaml model file. Data products are saved upon receipt of a tick or by command, and can be restored back into the data product database by command or at Set_Up. The store is protected by a CRC which is written on every save and checked prior to every restore, protecting against the restoration of corrupted or never-initialized memory contents. A command is provided to dump the current contents of the store into a packet. The autocoder limits the total size of the store to fit within a single Packet.T.
```
dinkelk: Maybe say <assembly_name>.stored_products.yaml to be more clear. I assume <name>.<assembly_name>.stored_products.yaml also works? Please verify, otherwise we will only be able to instantiate this component one time in a given assembly.

dinkelk: The idea of not restoring corrupted data makes sense. To be safe, I think we should also require that any data product included in this model is proven to have a type that is ALWAYS_VALID. this will prevent any errant constraint_error from taking down a system.

src/components/product_store/product_store.component.yaml
```diff
@@ -0,0 +1,52 @@
...
+      type: Product_Store_Types.Store_Description_Access_Type
+      description: The description of the data product store to manage.
+      not_null: true
+init:
+  description: The component is initialized by providing the memory region it is to manage, which holds the data product store.
+  parameters:
+    - name: Bytes
+      type: Basic_Types.Byte_Array_Access
+      not_null: true
+      description: A pointer to an allocation of bytes to be used for storing the data products. The size of this byte array MUST be at least Store_Description.Store_Size bytes in length. Only the first Store_Description.Store_Size bytes will be used by this component.
```
dinkelk: hopefully that includes the CRC.

src/components/product_store/product_store.commands.yaml
```diff
@@ -0,0 +1,9 @@
+---
+description: These are the commands for the Product Store component.
+commands:
+  - name: Save_Products
+    description: Save the configured data products from the database into the store. This performs the same operation as the receipt of a tick. If the store's save time is configured as Tick_Time, the current time is used instead, since no tick is available.
+  - name: Restore_Products
```
dinkelk, line 6: I am thinking two more commands might be useful
1) Enable_Save_On_Tick
2) Disable_Save_On_Tick

I feel like these could potentially be mission saving in some sort of crazy scenario where bad data is feeding the store and you need to override and disable the automatic save.

src/components/product_store/product_store.events.yaml
```diff
@@ -0,0 +1,24 @@
+---
+description: Events for the Product Store component.
+events:
+  - name: Products_Saved
+    description: The data products were saved to the store by command.
+  - name: Products_Restored
+    description: The data products held in the store were restored into the data product database.
+  - name: Store_Dumped
+    description: Produced a packet with the contents of the store.
+  - name: Data_Product_Missing_On_Save
```
dinkelk, line 16: This potentially could be spammy. One way we have dealt with this in Adamant is to have a boolean at init for the component that is Issue_Missing_On_Save event, so you could disable this feature when you know you potentially could be running for a long time without DPs in the database.

src/components/product_store/product_store.events.yaml
```diff
@@ -0,0 +1,24 @@
...
+    param_type: Invalid_Data_Product_Length.T
+  - name: Store_Crc_Invalid
+    description: The store CRC did not validate prior to a restore, so the restore was not performed. This is expected on the first boot before the store has ever been written.
+    param_type: Store_Crc_Error.T
+  - name: Invalid_Command_Received
+    description: A command was received with invalid parameters.
+    param_type: Invalid_Command_Info.T
+  - name: Dropped_Command
+    description: A command was dropped due to a full queue.
+    param_type: Command_Header.T
```
dinkelk, line 36: I am thinking that this component could benefit from some simple counter data products:
1) U16 roll over save count
2) U16 roll over restore count
3) U16 roll over CRC mismatch on restore count

src/components/product_store/test/test_assembly/test_assembly.stored_products.yaml
```diff
@@ -0,0 +1,16 @@
+---
+description: This is an example data product store used for testing the product store component.
+save_time: Tick_Time
+data_products:
+  - name: Test_Component_1_Instance.Data_Product_A
+    description: A data product stored with its timestamp and restored with that timestamp.
+    store_timestamp: True
+    restore_time: Use_Stored_Dp_Time
+    event_on_missing: True
```
dinkelk: ahh, i like this for event control. ignore my earlier comment in the events.yaml about this

src/components/product_store/test/test_assembly/backup.test_assembly.stored_products.yaml
```diff
@@ -0,0 +1,16 @@
...
+    store_timestamp: True
+    restore_time: Use_Stored_Dp_Time
+    event_on_missing: True
+  - name: Test_Component_1_Instance.Data_Product_B
+    description: A data product stored without its timestamp and restored with the save time.
+    store_timestamp: False
+    restore_time: Use_Save_Time
+    event_on_missing: False
+  - name: Test_Component_1_Instance.Data_Product_C
+    description: A data product stored without its timestamp and restored with a timestamp of zero.
```
dinkelk, line 13: Looking at this now, I am wondering why we are supporting saving WITHOUT a timestamp at all... will think more on this. I am not sure what the benefit is of this use case except maybe to save space in MRAM/memory dump

dinkelk: Actually, having second thoughts. Maybe the feature is OK. But my issue with this is that the default behavior here is to restore with a TS of zero, which i think is bad. The default should be Use_Save_Time in this case, since the save time is always available.

src/components/product_store/gen/schemas/stored_products.yaml
```diff
@@ -0,0 +1,59 @@
...
+  # Description of the data product store.
+  description:
+    type: str
+    required: False
+  # The source of the save time written to the store header on each save.
+  # This field can be set to:
+  #   Current_Time - The current time, fetched from the time connector, is written on each save.
+  #   Tick_Time - The time found on the incoming Tick.T is written on each save. If a save is
+  #               commanded, the current time is used instead.
+  #   No_Time - No save time is written to the store. The store header contains only the CRC.
```
dinkelk: Not sure if we should support No_Time here. A default of always saving either current or tick time seem reasonable, costs basically nothing, and only helps inform a dump.

src/components/product_store/gen/schemas/stored_products.yaml
```diff
@@ -0,0 +1,59 @@
...
+          name:
+            type: str
+            required: True
+          # Description of the entry.
+          description:
+            type: str
+            required: False
+          # Include this data product's timestamp just before its value in the store.
+          # By default this value is false.
+          store_timestamp:
```
dinkelk, line 40: If this is specified and restore_time is not Use_Stored_Dp_Time we should at least warn the user. Maybe disallow completely. I want to prevent a mistake by the user where they specify they want the DP TS saved but then accidentally tell restore to issue with a different timestamp. Let's dissallow for now until we have a use case that violates this.

src/components/product_store/types/store_crc_error.record.yaml
```diff
@@ -0,0 +1,15 @@
+---
+description: A packed record which holds data related to an invalid data product store CRC.
```
dinkelk: Is there already a type in adamant that is basically just this that we can reuse? If not, maybe we should make one.

src/components/product_store/types/product_store_types.ads
```diff
@@ -0,0 +1,66 @@
...
+
+-- Types used by the Product Store component to describe the set of data
+-- products that it saves and restores. Static instances of these types are
+-- autocoded from a stored_products.yaml model file.
+package Product_Store_Types is
+
+   -- The source of the save time written to the store header on each save:
+   type Save_Time_Type is (
+      -- No save time is written to the store. The store header contains only the CRC.
+      No_Time,
```
dinkelk: don't think we need this as feature

src/components/product_store/types/product_store_types.ads
```diff
@@ -0,0 +1,66 @@
...
+   );
+
+   -- Description of a single data product entry in the store:
+   type Store_Entry_Type is record
+      -- Data product identifier of the entry:
+      Data_Product_Id : Data_Product_Types.Data_Product_Id := Data_Product_Types.Data_Product_Id'First;
+      -- Should the data product's timestamp be saved to the store just before its value?
+      Store_Timestamp : Boolean := False;
+      -- The timestamp to apply to this data product when it is restored:
+      Restore_Time : Restore_Time_Type := Use_Zeros;
```
dinkelk: I don't like this as a default.

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+         -- An entry may only be restored with the stored data product time if that
+         -- time is actually stored:
+         pragma Assert (Item.Restore_Time /= Use_Stored_Dp_Time or else Item.Store_Timestamp);
+         -- An entry may only be restored with the save time if a save time is
+         -- actually stored:
+         pragma Assert (Item.Restore_Time /= Use_Save_Time or else Self.Store_Description.Save_Time /= No_Time);
+      end loop;
+      -- The store size found in the description must match the size computed from
+      -- the entries:
+      pragma Assert (Expected_Size = Self.Store_Description.Store_Size);
```
dinkelk, line 84: assertions here that the dump size fits within packet.t payload might be nice too, unless you have that somewhere else. this pattern exists elsewhere in adamant if you want to lookl

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+      Self.Commands_Dispatched_Per_Tick := Commands_Dispatched_Per_Tick;
+   end Init;
+
+   ---------------------------------------
+   -- Private helper subprograms:
+   ---------------------------------------
+
+   -- Return the index of the first byte of the store data region, which is the
+   -- region covered by the CRC:
+   function Data_First (Self : in Instance) return Natural is
```
dinkelk: with Inline => True might be a good call on all these 1 liners

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+   -- Save the data products into the store, stamping the store with the provided
+   -- save time:
+   procedure Do_Save (Self : in out Instance; Save_Time : in Sys_Time.T) is
+      use Basic_Types;
+      use Data_Product_Enums.Fetch_Status;
+      -- If the store currently holds valid contents, then slots for data products
+      -- that cannot be fetched are left untouched, preserving the last saved
+      -- values. Otherwise the slots are zeroed, so that stale memory contents can
+      -- never be presented as valid data.
+      Store_Was_Valid : constant Boolean := Self.Compute_Store_Crc = Self.Read_Stored_Crc;
```
dinkelk: I don't understand this at all. We are getting ready to overwrite the entire store with new data. Why do we need to compute CRC prior and check anything at all?

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+                  -- Check the length of the data product to make sure it is what we expect:
+                  if Fetch_Return.The_Data_Product.Header.Buffer_Length /= Item.Size then
+                     Self.Event_T_Send_If_Connected (Self.Events.Data_Product_Length_Mismatch (Self.Sys_Time_T_Get, (
+                        Header => Fetch_Return.The_Data_Product.Header,
+                        Expected_Length => Item.Size)
+                     ));
+                  else
+                     Save_Slot := True;
+                  end if;
+               when Not_Available | Id_Out_Of_Range =>
```
dinkelk: ID out of range is a worse error and should not be gated by your boolean. In this case we should unconditionally throw a different event, specific to this case.

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+                  Write_Idx : Natural := Idx;
+               begin
+                  if Item.Store_Timestamp then
+                     Self.Bytes.all (Write_Idx .. Write_Idx + Time_Length - 1) := Sys_Time.Serialization.To_Byte_Array (Fetch_Return.The_Data_Product.Header.Time);
+                     Write_Idx := @ + Time_Length;
+                  end if;
+                  Self.Bytes.all (Write_Idx .. Write_Idx + Item.Size - 1) :=
+                     Fetch_Return.The_Data_Product.Buffer (Fetch_Return.The_Data_Product.Buffer'First .. Fetch_Return.The_Data_Product.Buffer'First + Item.Size - 1);
+               end;
+            elsif not Store_Was_Valid then
```
dinkelk: I don't see why this is necessary.

src/components/product_store/gen/schemas/stored_products.yaml
```diff
@@ -0,0 +1,59 @@
...
+  # The source of the save time written to the store header on each save.
+  # This field can be set to:
+  #   Current_Time - The current time, fetched from the time connector, is written on each save.
+  #   Tick_Time - The time found on the incoming Tick.T is written on each save. If a save is
+  #               commanded, the current time is used instead.
+  #   No_Time - No save time is written to the store. The store header contains only the CRC.
+  # By default, if not specified, this value is set to Current_Time.
+  save_time:
+    type: str
+    enum: ['Current_Time', 'Tick_Time', 'No_Time']
```
dinkelk: Another interesting option here would be "Data_Product_Time". And if specified, the a data product in the list below must specify:

```
store_timestamp: Use_As_Global
```
or something similar. There might be a better way to achieve it. Could use a new tag like `use_timestamp_as_global: True`

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+      end loop;
+
+      -- Compute the CRC over the store contents and write it to the header:
+      Self.Bytes.all (Self.Bytes.all'First .. Self.Bytes.all'First + Crc_Length - 1) := Self.Compute_Store_Crc;
+   end Do_Save;
+
+   -- Restore the data products held in the store into the data product database.
+   -- Status is set to True if the restore was performed, or False if the store CRC
+   -- did not validate.
+   procedure Do_Restore (Self : in out Instance; Status : out Boolean) is
```
dinkelk: why not return status as function return instead of out param?

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+      Status := True;
+   end Do_Restore;
+
+   -- Build and send a packet containing the contents of the store:
+   procedure Do_Dump (Self : in out Instance) is
+      use Serializer_Types;
+      Pkt : Packet.T;
+      Stat : constant Serialization_Status := Self.Packets.Stored_Products (Self.Sys_Time_T_Get, Self.Bytes.all (Self.Bytes.all'First .. Self.Store_Last), Pkt);
+   begin
+      -- This should never fail since both the autocoder and an assertion at Init
```
dinkelk: is there such an assertion at Init?

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+   -- Invokee connector primitives:
+   ---------------------------------------
+   -- This is the base tick for the component. Each tick received saves the data
+   -- products to the store.
+   overriding procedure Tick_T_Recv_Sync (Self : in out Instance; Arg : in Tick.T) is
+      Messages_Dispatched : Natural;
+   begin
+      -- Handle any commands in the queue. Service up to N commands per tick:
+      Messages_Dispatched := Self.Dispatch_N (Self.Commands_Dispatched_Per_Tick);
+      pragma Assert (Messages_Dispatched <= Self.Commands_Dispatched_Per_Tick);
```
dinkelk, line 496: Hm. I am not sure I like command handling done here. What if the tick rate is really low. Like 1x every 10 minutes. This would mean that we can only execute commands every 10 minutes.

To get around this we may want to set up an initial divider at Init. So we could run this at something reasonable for commands, like 1 Hz, but then slow saving down to 10 minutes using a Ticks_Per_Save parameter (default 1) at 600.

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+   ---------------------------------------
+   -- Invokee connector primitives:
+   ---------------------------------------
+   -- This is the base tick for the component. Each tick received saves the data
+   -- products to the store.
+   overriding procedure Tick_T_Recv_Sync (Self : in out Instance; Arg : in Tick.T) is
+      Messages_Dispatched : Natural;
+   begin
+      -- Handle any commands in the queue. Service up to N commands per tick:
+      Messages_Dispatched := Self.Dispatch_N (Self.Commands_Dispatched_Per_Tick);
```
dinkelk, line 495: This dispatch should be done after the business logic.

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+   -- performs the same operation as the receipt of a tick. If the store's save time
+   -- is configured as Tick_Time, the current time is used instead, since no tick is
+   -- available.
+   overriding function Save_Products (Self : in out Instance) return Command_Execution_Status.E is
+      use Command_Execution_Status;
+   begin
+      -- Save using the current time as the save time. Note that if the store is
+      -- configured for Tick_Time, the current time is used for a commanded save,
+      -- since no tick is available:
+      case Self.Store_Description.Save_Time is
```
dinkelk: code duplication here with the tick. consolidate.

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,361 @@
...
+   end Save_Products;
+
+   -- Restore the data product values held in the store back into the data product
+   -- database. The store CRC is checked prior to the restore, and the command fails
+   -- if the CRC does not validate.
+   overriding function Restore_Products (Self : in out Instance) return Command_Execution_Status.E is
+      use Command_Execution_Status;
+      Restore_Ok : Boolean;
+   begin
+      Self.Do_Restore (Restore_Ok);
```
dinkelk: just use function with return boolean instead to clean this up;

dinkelk (changes requested):

src/util/crc/crc_mismatch_info.record.yaml
```diff
@@ -0,0 +1,15 @@
...
+    type: Crc_16.Crc_16_Type
+    format: U16
+    skip_validation: True
+    byte_image: True
+  - name: Expected_Crc
+    description: The CRC that the computed CRC was expected to match.
+    type: Crc_16.Crc_16_Type
+    format: U16
+    skip_validation: True
+    byte_image: True
```
dinkelk, line 15: I think i'd rather this live with src/util/crc/. One less dir to add to the path for build performance, and it makes some sense to cluster it with the utility.

src/components/product_store/gen/models/stored_products.py
```diff
@@ -337,6 +343,26 @@ def _resolve_data_products(self, assembly):
...
             entry.size = entry.data_product.type_model.size  # in bits
 
+            # Make sure the data product type is "always valid", meaning that no
+            # bit representation of the type can fail validation. The store zeroes
+            # the slots of data products that cannot be fetched on save, and MRAM
+            # contents could theoretically be corrupted without violating the CRC
+            # (i.e. a matching recomputation). Requiring always valid types
+            # guarantees a restore can never inject a data product whose use
+            # downstream raises a constraint error:
+            if not entry.data_product.type_model.is_always_valid():
```
dinkelk, line 389: This is good. We should belt and suspenders it though. Add a compile time error to the generated Ada package which prevents compilation if somehow a possibly invalid type sneaks in.

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -146,9 +151,8 @@ package body Component.Product_Store.Implementation is
                   Self.Bytes.all (Write_Idx .. Write_Idx + Item.Size - 1) :=
                      Fetch_Return.The_Data_Product.Buffer (Fetch_Return.The_Data_Product.Buffer'First .. Fetch_Return.The_Data_Product.Buffer'First + Item.Size - 1);
                end;
-            elsif not Store_Was_Valid then
-               -- The data product could not be saved and the existing slot contents
-               -- are not trustworthy, so zero the slot:
+            else
+               -- The data product could not be saved, so zero the slot:
```
dinkelk: What is the logic behind this? Intuitively to me, if we have issue fetching the data product, we should do nothing. Leave whatever the last value that was good was from the store. Overwriting with zeros could mean anything, and seems better to keep the old value.

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -159,35 +163,38 @@ package body Component.Product_Store.Implementation is
...
-      if Self.Store_Description.Save_Time /= No_Time then
-         Save_Time_Stamp := Sys_Time.Serialization.From_Byte_Array (Self.Bytes.all (Idx .. Idx + Time_Length - 1));
-         Idx := @ + Time_Length;
-      end if;
+      -- Read the save time:
+      Save_Time_Stamp := Sys_Time.Serialization.From_Byte_Array (Self.Bytes.all (Idx .. Idx + Time_Length - 1));
+      Idx := @ + Time_Length;
 
       -- Restore each data product entry:
       for Item of Self.Store_Description.Entries.all loop
```
dinkelk: Unless I am missing something, I think we have an issue here. What if a data product was never stored. Whatever is in MRAM should not be restored to the database. Ideally, we would do nothing in that case, and leave the data product as Not_Available in the database.

The issue is that I think we would need some meta data to indicate that a DP is written, which doesn't currently exist. In the store you could include a byte for each DP that you write a zero or nonzero value for to indicate if the DP has been saved or never saved. We can trust that this store is not garbage if we CRC over it, so that if the CRC is good, we know this meta data store is meaningfull. Let's chat before you implement this. There could be better ideas.

src/components/product_store/gen/schemas/stored_products.yaml
dinkelk: Let's pass on this for now. If there is a use case later, we can do this.

dinkelk (changes requested):

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,399 @@
...
+      -- Compute the expected size of the store from the description, and check the
+      -- configuration of each entry:
+      for Item of Self.Store_Description.Entries.all loop
+         Expected_Size := @ + Item.Size;
+         if Item.Store_Timestamp then
+            Expected_Size := @ + Time_Length;
+         end if;
+         -- An entry stores its data product's timestamp if and only if that
+         -- timestamp is what the entry is restored with:
+         pragma Assert (Item.Store_Timestamp = (Item.Restore_Time = Use_Stored_Dp_Time));
```
dinkelk, line 80: If this is true, do we need Store_Timestamp boolean at all? Could we just use Restore_Time to do the logic we need?

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,399 @@
...
+
+   -- Build and send a packet containing the contents of the store:
+   procedure Do_Dump (Self : in out Instance) is
+      use Serializer_Types;
+      Pkt : Packet.T;
+      Stat : constant Serialization_Status := Self.Packets.Stored_Products (Self.Sys_Time_T_Get, Self.Bytes.all (Self.Bytes.all'First .. Self.Store_Last), Pkt);
+   begin
+      -- This should never fail since the autocoder and the Store_Size_Type
+      -- constraint guarantee that the store fits within a single packet:
+      pragma Assert (Stat = Success);
```
dinkelk, line 426: I still don't see an assertion Init that proves this comment. Where is it?

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,399 @@
...
+      -- tick divider has elapsed:
+      if Self.Save_On_Tick and then Self.Tick_Count = 0 then
+         Self.Do_Save (Save_Time =>
+            (case Self.Store_Description.Save_Time is
+               when Tick_Time => Arg.Time,
+               when Current_Time => Self.Sys_Time_T_Get));
+      end if;
+
+      -- Increment the tick counter, rolling over at Ticks_Per_Save:
+      Self.Tick_Count := (@ + 1) mod Self.Ticks_Per_Save;
```
dinkelk: Seems like this should be gated by Save_On_Tick to be more correct and save computation if Save_On_Tick is False.

src/components/product_store/component-product_store-implementation.adb
```diff
@@ -0,0 +1,399 @@
...
+      return Success;
+   end Dump_Store;
+
+   -- Enable the automatic saving of the data products to the store upon receipt of
+   -- a tick.
+   overriding function Enable_Save_On_Tick (Self : in out Instance) return Command_Execution_Status.E is
+      use Command_Execution_Status;
+   begin
+      Self.Save_On_Tick := True;
+      Self.Event_T_Send_If_Connected (Self.Events.Save_On_Tick_Enabled (Self.Sys_Time_T_Get));
```
dinkelk, line 566: Seems like we should reset count related to this here.

src/components/product_store/component-product_store-implementation.adb
dinkelk: Maybe we want this, just in case we want to store the DP timestamp for dump. Weird use case, but maybe OK.

dinkelk (changes requested):

dinkelk:
One somewhat large design issue that is not addressed:

No atomicity across a save. Do_Save overwrites the store in place and writes the CRC last. A reboot mid-save corrupts the entire store, so the Set_Up restore fails and all products are lost, not just the ones being written. Power loss during a write is fairly central to this component's mission. The CRC-last ordering means it fails safe, which is good, but a double-buffered layout (two copies, alternate writes, restore the valid/newer one) would survive it.

We should discuss this as an option. AI suggests:

```
Ping-pong with a sequence number (double buffer done properly). Worth distinguishing from a naive double buffer: two copies, alternate writes, and a monotonic save
  counter inside each CRC-protected region. Restore validates both, picks the valid copy with the newer counter. You never overwrite the only good copy, and a mid-save
  reboot costs you exactly one save interval of freshness. This is the classic EEPROM/MRAM pattern and the most robust option, at the cost of doubling the memory footprint
  and complicating the dump command (dump which copy? probably both, or the active one plus a data product reporting which is active).
```

Alternatively, we go to a best effort mode instead, and potentially drop the CRC altogether.

dinkelk:
A few AI flagged issues worth looking at:

```
  2. component-product_store-implementation.ads doc drift: the Init comment block lists Bytes, Restore_On_Set_Up, and Commands_Dispatched_Per_Tick but omits Ticks_Per_Save
  (the .adb version has all four). Also the private-part connector comment still says "Each tick received saves the data products to the store," which predates the divider
  and command dispatch.
  3. Dead/hardcoded code in gen/models/stored_products.py: crc_obj = [None] is declared and never used, and _get_crc_size() hardcodes 16 bits rather than deriving it from
  the Crc_16 type. The hardcode is acknowledged in a comment and probably fine, but the dead variable should go.
  4. Unhelpful failure mode in gen/models/product_store_packets.py: in set_assembly, if the specific-name path matching finds nothing, model_path stays None and
  model_loader.load_model(None) will blow up with an obscure error. An explicit ModelException naming the model it looked for would save someone an afternoon. Related: the
  configs.split(".")[0].split("_Stored_Products") parsing is case-sensitive, but Ada is not, so a user writing the discriminant as
  TEST_ASSEMBLY_STORED_PRODUCTS.Store_Description'Access would compile fine and fail model generation confusingly.
6. No test passes a byte array larger than Store_Size or one with a nonzero 'First, both of which Init explicitly permits. The index arithmetic (Data_First, Store_Last)
  handles both correctly as written, but a test would pin that behavior against refactors.
  7. Test_Missing_Data_Product verifies the zeroed store bytes but never restores them, so the "zeros flow back into the database" behavior is untested. If the validity-byte
  redesign happens, this test becomes the natural home for "invalid slots are skipped on restore."
  ```

PR #197, opened 2026-07-31
Make Sys_Time.Assertion safe on coarse-tick runtimes

dinkelk:
## Summary

`Sys_Time_Assert.Eq`/`Neq` raise `CONSTRAINT_ERROR` on every call on embedded cross targets, even when the compared times are equal. Found via a renode (RV32) unit test in adamant-xmera-components (lasp/adamant-xmera-components#43): the first test there to call `Sys_Time_Assert.Eq` on the cross target died with an unexplained `CONSTRAINT_ERROR`.

Two defects, both fixed:

1. **Eager message construction.** The diagnostic text was built as a declaration, so its arithmetic ran even when the assertion passed. The text is now constructed only in the failure branch (the assert call itself is preserved on both paths so AUnit's assertion accounting is unchanged).

2. **Target-dependent unit conversion.** The text rendered the difference as `(T1 - T2) / Nanoseconds (1)`. Per the RM, `Nanoseconds` rounds to the nearest multiple of `Time_Span_Unit`, so on any runtime with a tick coarser than 1 ns it returns a zero span and the division raises `CONSTRAINT_ERROR`. (`Eps / Nanoseconds (1)` on the next line had the same bug.) The difference and tolerance are now rendered in seconds via `Duration'Image (To_Duration (...))`, which constructs no `Time_Span` and cannot overflow for any representable span — unlike an integer-nanoseconds rendering via `To_Duration (...) * 1e9`, which overflows for differences over ~9.2 s.

Additionally, the default tolerance is now derived from the type instead of the bare `Nanoseconds (300)`: the largest of

- **half a subsecond LSB** — differences below this are representational rounding of `Sys_Time.T` itself,
- **300 ns** — the historical conversion-error allowance, preserved as a floor so no existing test gets a tighter default,
- **one `Time_Span_Unit`** — so the tolerance never rounds to zero on coarse-tick targets (previously the 300 ns default silently degraded to exact-match there).

For the base (1/2^32 subseconds) variant the default is unchanged at 300 ns. For subseconds_16 it widens to half a tick (~7.6 µs), which only affects `Eq` in the passing direction; the repo has no `Neq` call sites relying on the default.

Both variants' spec/body files remain byte-identical to each other.

### DIFF
--- a/src/types/sys_time/sys_time-assertion.adb
+++ b/src/types/sys_time/sys_time-assertion.adb
@@ -1,5 +1,4 @@
 with Sys_Time.Arithmetic;
-with Delta_Time.Arithmetic;
 
 -- Smart assert for comparing system times
 -- Useful when you need to use the >, >=, <, and <= operators.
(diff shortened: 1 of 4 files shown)


src/types/sys_time/sys_time-assertion.adb
```diff
@@ -24,26 +23,37 @@ package body Sys_Time.Assertion is
       end if;
    end Sys_Time_Equal;
 
+   -- Diagnostic text for a failed comparison. The difference and tolerance
+   -- are rendered in seconds via To_Duration, which is safe on any target:
+   -- it never constructs a Time_Span from a unit finer than the runtime
+   -- tick. (The previous rendering divided by Nanoseconds (1), which rounds
+   -- to a zero Time_Span on runtimes with a tick coarser than a nanosecond,
+   -- raising Constraint_Error.)
```
dinkelk: this comment is already stale. please go through this PR and try to avoid stale comments that mention this issue that will have been solved as soon as this is merged.

dinkelk: Fixed in a4b37b4d. Reworded as a live constraint (do not render by dividing by a constructed Time_Span) and also removed the change-relative wording from the Default_Eps comment in the specs — no other stale references found in the diff.

Jbsco (approved):

PR #198, opened 2026-08-05
Convert the Stopwatch packages to SPARK, proven at gold level

dinkelk:
## Summary

Converts the `Stopwatch` and `Stopwatch.Reporter` packages (plus the one `Delta_Time.Arithmetic` subprogram they call) to SPARK and proves them at **gold level — flow analysis, absence of runtime errors, and functional contracts — with `--checks-as-errors=on` and zero unproved checks** (64 checks total; 20+ functional contracts discharged).

These packages time every rate group in every Adamant assembly, so their correctness is safety-relevant; the proof now guarantees no runtime errors and that the measurement bookkeeping and max-folding math do exactly what the contracts say. Downstream packages can build their own proofs on top of these contracts, reusing the ghost predicates across the package boundary.

## API and behavior changes

- `Reporter.Reset` takes `Self` as `out` instead of `in out` — it overwrites the whole instance, and flow analysis flags the false data dependency. Callers are unaffected; `out` additionally permits passing an uninitialized instance, so Reset doubles as an initializer.
- `Delta_Time.Arithmetic.To_Delta_Time` gains a procedure form (a function with an `out` parameter is not callable from SPARK) and is made robust while here: the output is fully assigned on every path, and a failed conversion now **saturates** (the previous behavior left the output unassigned).
- Everything else is contract-only: `SPARK_Mode`, `Global`, and `Post` have no effect on generated code with assertions disabled, and the contracts hold at runtime when enabled.

## Commit 1 — SPARK conversion

- `SPARK_Mode => On` with `Global` data-dependency contracts on every subprogram; timer starts/stops declare their reads of `Ada.Real_Time.Clock_Time` (and `Ada.Task_Identification.Tasking_State` via `Current_Task` for CPU timers).
- Volatile clock reads hoisted into constants (the only placement SPARK allows); `Current_Task` hoisted explicitly since evaluating it as `Clock`'s default parameter is an interfering context (SPARK RM 7.1.3(9)). The pattern is consolidated in one private volatile helper, `Cpu_Now`, with a documented `pragma Assume` discharging Clock's non-null-task precondition.
- One justified check: the wall `Result` subtraction's range check, with the full argument in the justification string (64-bit `Time`/`Time_Span`, no wider type, ~292-year overflow horizon for same-clock stamps).

## Commit 2 — functional contracts (gold)

- Start/Stop contracts pin which timestamps are preserved and that stored measurements equal the timer `Result`s.
- `Accumulate` states the max-folding math exactly via the `Maximums_Accumulated` ghost predicate shared by both overloads; the exceeded flags fire precisely when an all-time max is beaten.
- `Reset`/`Reset_Recent_Max` prove what is zeroed and what is preserved via `Is_Reset` and `Recent_Max_Reset`; downstream proofs reuse these predicates across the package boundary.
- Ghost predicate declarations sit with the public interface (they are the vocabulary the contracts are written in); their expression-function definitions live in the private part. Ghost code compiles to nothing.

### DIFF
--- a/src/types/sys_time/arithmetic/delta_time-arithmetic.adb
+++ b/src/types/sys_time/arithmetic/delta_time-arithmetic.adb
@@ -18,6 +18,11 @@ package body Delta_Time.Arithmetic is
       -- Convert from time to Sys_Time
       Status : Sys_Time_Status;
    begin
+      -- Give the output a defined zero value up front so that it is fully
+      -- assigned on every path, including the Constraint_Error handler
+      -- below:
+      Arg_Out := Zero_Delta_Time;
+
       if Arg_In = Time_Span_First then
          -- Special case for Time_Span_First, which would overflow abs if made positive. In this
          -- case we round to the nearest positive time span value.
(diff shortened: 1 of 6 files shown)


Jbsco (approved):
Nice SPARK conversion! ⚡

PR #200, opened 2026-08-10
Add prove_all rule and workflow to prove all SPARK packages

Jbsco:
Adds `prove_all`, the recursive prove counterpart to `test_all`/`coverage_all`, and workflows that run it on push/PR to `main` plus an arm64 variant on release.

It walks the repo and runs `redo prove` in every directory with an `all.prove.yaml` or a `SPARK_Mode => On` source (only `gen/templates` is excluded, as in `style_all`), so SPARK packages are proved automatically as they're added. Nothing is hardcoded, and SPARK inside autocode is proved through its owning directory. A directory can opt out with `.skip_prove`.

- `redo/rules/build_prove_all.py` + `default.do` / `build_what_predefined.py` wiring
- `.github/workflows/prove_all.yml` and `.github/workflows/prove_all_arm64.yml`
- `.skip_prove` on the `spark_compile` proof-failure fixture

Today it discovers and proves `src/util/stopwatch`, the two `doc/example_architecture` SPARK examples, and `redo/test/aspect_compile`.

### DIFF
--- a/redo/rules/build_prove_all.py
+++ b/redo/rules/build_prove_all.py
@@ -1,0 +1,155 @@
+import os.path
+import re
+import sys
+from util import redo
+from util import error
+from util import filesystem
+from base_classes.build_rule_base import build_rule_base
+from shutil import copytree
+
+# Definitions for producing colored text on the terminal
+NO_COLOR = "\033[0m"
+BOLD = "\033[1m"
+RED = "\033[31m"
+GREEN = "\033[32m"
+if "REDO_PROVE_ALL_NO_COLOR" in os.environ and os.environ["REDO_PROVE_ALL_NO_COLOR"]:
+    PASSED = "PASSED"
+    FAILED = "FAILED"
+else:
+    PASSED = BOLD + GREEN + "PASSED" + NO_COLOR
+    FAILED = BOLD + RED + "FAILED" + NO_COLOR
+
+SPARK_ON_REGEX = re.compile(r"SPARK_Mode\s*=>\s*On")
+
+
+def _dir_has_spark_on(files, root):
+    """Return True if any Ada source in this directory enables SPARK."""
+    for f in files:
+        if f.endswith(".ads") or f.endswith(".adb"):
+            try:
+                with open(os.path.join(root, f), "r", errors="ignore") as fh:
+                    if SPARK_ON_REGEX.search(fh.read()):
+                        return True
+            except OSError:
+                pass
+    return False
+
+
+class build_prove_all(build_rule_base):
+    """
+    This build rule looks for SPARK packages in the directory it is
+    passed and recursively below. A directory is eligible if it contains
+    an "all.prove.yaml" GNATprove configuration or any source with
+    "SPARK_Mode => On". It then runs "redo prove" in each eligible
+    directory in sequence and prints a report to the terminal as the
+    proofs are run. Eligibility is recomputed from the source tree on
+    every run, so a package converted to SPARK is picked up
+    automatically. Note that "redo prove" analyzes a directory's full
+    object closure, including generated code, so SPARK inside autocode
+    is proved through its owning directory.
+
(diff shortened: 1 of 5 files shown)


src/util/stopwatch/all.prove.yaml
```diff
@@ -0,0 +1,4 @@
+---
+description: GNATprove configuration for the Stopwatch SPARK packages
+level: 2
+mode: "gold"
```
dinkelk: Why was this added?

redo/rules/build_prove_all.py
```diff
@@ -0,0 +1,129 @@
...
+    FAILED = "FAILED"
+else:
+    PASSED = BOLD + GREEN + "PASSED" + NO_COLOR
+    FAILED = BOLD + RED + "FAILED" + NO_COLOR
+
+# Directories never eligible for prove-all: build artifacts, alire crates,
+# unit-test fixtures (which include deliberate proof-failure cases such as
+# redo/test/spark_compile/spark_fail), documentation examples, and code
+# generator templates (placeholder units, not real source).
+IGNORE_DIRS = ["build", "alire", "test", "doc", "gen"]
```
dinkelk: Ignoring build is not a good idea. We might and should start putting SPARK inside of autocode.

redo/rules/build_prove_all.py
```diff
@@ -0,0 +1,129 @@
...
+if "REDO_PROVE_ALL_NO_COLOR" in os.environ and os.environ["REDO_PROVE_ALL_NO_COLOR"]:
+    PASSED = "PASSED"
+    FAILED = "FAILED"
+else:
+    PASSED = BOLD + GREEN + "PASSED" + NO_COLOR
+    FAILED = BOLD + RED + "FAILED" + NO_COLOR
+
+# Directories never eligible for prove-all: build artifacts, alire crates,
+# unit-test fixtures (which include deliberate proof-failure cases such as
+# redo/test/spark_compile/spark_fail), documentation examples, and code
```
dinkelk: redo/test/spark_compile/spark_fail should just contain a skip file no?

dinkelk (changes requested):
Other things: 

1. I think we need an arm variant. This ensures that the tools work on both platforms we support. The justification in the PR description is not good enough. The same could be said for style, documentation, etc.
2. Why would we not allow SPARK in test or build?

src/util/stopwatch/all.prove.yaml
Jbsco: Removed

redo/rules/build_prove_all.py
Jbsco: Build is only ignored for marker scanning, SPARK inside autocode still gets proved. `redo <dir>/prove` runs on the full object closure including build/src generated units.
test, doc, gen category exclusions removed.

redo/rules/build_prove_all.py
Jbsco: Agreed, added

redo/rules/build_prove_all.py
```diff
@@ -0,0 +1,157 @@
...
+    def build(self, redo_1, redo_2, redo_3):
+        import database.setup
+
+        # Figure out build directory location
+        directory = os.path.abspath(os.path.dirname(redo_1))
+
+        # Find all SPARK directories below this directory:
+        prove_dirs = []
+        for root, dirnames, files in filesystem.recurse_through_repo(directory):
+            if os.sep + "alire" + os.sep in root:
```
dinkelk: trailing os.sep  needed? seems weird to me

dinkelk (changes requested):
PR description needs update

redo/rules/build_prove_all.py
Jbsco: Removed

dinkelk (approved):

PR #202, opened 2026-08-12
Support named pad bytes in the product packetizer packet model

dinkelk:
## Summary

This PR allows a `pad_bytes` entry in a product packets model to optionally carry a `name`. The name labels the reserved space in the generated packet layout and downstream ground artifacts (HTML packet documentation, COSMOS plugin telemetry definitions, Hydra configuration) instead of the auto-generated `Reserved_NN` placeholder. This is useful when a packet reserves space for telemetry that does not exist yet: operators and downstream tools can reference the reserved field by its intended name today, instead of an anonymous placeholder documented only in YAML comments.

```yaml
data_products:
  - name: Omeg_Norm                                # named pad
    pad_bytes: 8
  - name: Some_Component_Instance.Future_Item      # named pad with a dotted name
    pad_bytes: 4
  - pad_bytes: 2                                   # anonymous pads, same as before
  - pad_bytes: 1
```

which renders in the COSMOS plugin as:

```
  Append_Array_Item Omeg_Norm 8 UINT 64 "8 reserved pad bytes for future item Omeg_Norm."
  Append_Array_Item Some_Component_Instance.Future_Item 8 UINT 32 "4 reserved pad bytes for future item Some_Component_Instance.Future_Item."
  Append_Array_Item Reserved_00 8 UINT 16 "2 pad bytes."
  Append_Item Reserved_01 8 UINT "1 pad byte."
```

Details:

- A pad name may be a bare identifier or a dotted path with any number of segments. A dotted name lets a pad carry the exact name the future real item will have in the ground dictionary, so ground references stay stable when the pad is later replaced by the real data product. Pad names are opaque labels: they are never resolved against the assembly.
- Generated pad descriptions state the byte count, and named pads additionally state the future item name.
- Pad items display as `Basic_Types.Byte_Array` (`Basic_Types.Byte` for a 1-byte pad), matching their `U8xN` format, instead of `Natural`. The type is display metadata only and is not used for serialization anywhere.
- Validation (all raise `ModelException`): `pad_bytes` combined with the behavioral options is still rejected, invalid or empty name segments are rejected, duplicate pad names within a packet are rejected, and a pad name that collides with a data product entry name in the same packet is rejected.
- Every packet item insertion (data product fields, included timestamps, pads) is now checked for name uniqueness within the packet, in any ordering. Colliding names previously either silently overwrote one another in the items dict or emitted duplicate item names into the ground artifacts, letting the ground definition diverge from the packet the flight software builds. This also rejects listing the same data product twice in one packet, which previously generated broken ground artifacts silently. The items dict is now keyed uniformly by packet-prefixed full name; the keys are internal, so generated artifacts are unchanged for all existing models.
- The generated runtime Ada is functionally unchanged: pad entries still have no data product ID, only a size. Only the item entry comments in the generated packet table change.
- Anonymous pads behave exactly as before, including `Reserved_NN` numbering.
- The generator documentation PDF is regenerated (it embeds the updated schema and test assembly example).

A third pre-existing defect in the same area is also fixed: the timestamp items inserted by `include_timestamp` were built with no container entity, so they rendered in the COSMOS plugin as a bare `.Seconds` / `.Subseconds` with an empty prefix, and two included timestamps in one packet would have produced colliding item names. They are now named after their data product (e.g. `Component_Instance.Data_Product.Seconds`), consistent with the value items.

Two pre-existing display defects in the flattened packet HTML documentation are also fixed: pad items rendered with blank start and stop bits, and any item following a pad rendered with offsets that ignored the pad size, because the pad branch never advanced the running bit position. Included timestamps had the same defect (the bit position restarted at zero), so a timestamp included for any item other than the first corrupted the offsets of everything after it. Both were display-only: the runtime packet layout and the COSMOS definitions are sequential and were unaffected. Pad items now carry proper packet start and stop bits, and the bit position threads through pads and included timestamps. The padding test packet now includes a mid-packet timestamp after a pad to cover both paths.

The first commit fixes a pre-existing bug found while testing: COSMOS telemetry generation fails on a 1-byte pad. A 1-byte pad has format `U8` with no array suffix, so it fell through to the scalar template branch, which dereferences `item.entity`, an attribute pad items do not have. The template now branches on the format object's `length` attribute instead of indexing individual characters of the format string, and renders entity-less items by their own name and description. This also fixes array formats with multi-digit unit sizes (e.g. `U16x4`), where single-character indexing produced a wrong element size and a total size of zero.


src/components/product_packetizer/gen/models/product_packets.py
```diff
@@ -60,25 +60,48 @@ def __init__(
...
             or include_timestamp
             or used_for_on_change is not True
         ):
             raise ModelException(
                 ('Packet data product list specifies "pad_bytes". If "pad_bytes" is specified'
-                 'then [name, event_on_missing, use_timestamp, include_timestamp, used_for_on_change] may not be specified.')
+                 ' then [event_on_missing, use_timestamp, include_timestamp, used_for_on_change] may not be specified.')
+            )
+
+        if description and not pad_bytes:
```
dinkelk: Huh? Seems like we should allow description for non pad_byte items as well? Have we not been allowing that before?

src/components/product_packetizer/gen/models/product_packets.py
```diff
@@ -267,23 +316,52 @@ def create_item_list(self):
                 # Update the items dict:
                 self.items.update(OrderedDict(zip(new_names, product_packet_items)))
             else:
+                # Determine the pad name and description. Named pads carry their
+                # user-provided label into the generated packet layout and ground
+                # artifacts. Anonymous pads are auto-named Reserved_NN as before.
```
dinkelk: no stale comments. "as before" is not an ok thing to comment.

src/components/product_packetizer/gen/models/product_packets.py
```diff
@@ -267,23 +316,52 @@ def create_item_list(self):
...
+                if dp.pad_name:
+                    item_name = dp.pad_name
+                    item_description = (
+                        dp.pad_description
+                        if dp.pad_description
+                        else "Reserved pad bytes for future item " + dp.pad_name + "."
```
dinkelk: Can we say how many bytes in the description here?

src/components/product_packetizer/gen/models/product_packets.py
```diff
@@ -267,23 +316,52 @@ def create_item_list(self):
...
+                    item_name = "Reserved_%02d" % pad_count
+                    pad_count += 1
+                    item_description = (
+                        dp.pad_description if dp.pad_description else "Pad bytes."
+                    )
+                    flat_description = (
+                        dp.pad_description
+                        if dp.pad_description
+                        else "Unused pad bytes."
+                    )
```
dinkelk: Same with this path. Number of bytes would be useful.

src/components/ccsds_packetizer/gen/templates/assembly/name_ccsds_cosmos_telemetry.txt
```diff
@@ -16,7 +16,8 @@ Telemetry {{ name }} {{ packet.name }} Big_Endian "{{ packet.description }}"
 {# The packet definitions. #}
 {% if packet.items %}
 {% for item_name, item in packet.items.items() %}
-  {% if 'x' in (item.format|string)[2] %}Append_Array_Item {% if item.type_model.name %}{{ item.type_model.name }}.{% endif %}{{ item.name }} {{ (item.format|string)[1] }} {{ plugin_format_dictionary[item.format.type[0]] }} {{ (item.format|string)[1]|int * (item.format|string)[3:]|int }} "{{ item.description|replace('\n','->') }}"
```
Jbsco, line 19: Dang - Thanks for catching this!

Jbsco (approved):

PR #206, opened 2026-08-18
Validate each component's default parameter values at assembly startup

cbrisebois-lasp:
A component's default parameter values are compiled in as record initializers and never pass through the component's own Validate_Parameters function, which runs only on ground-commanded table operations, so a default violating the component's cross-parameter checks flies silently until a table upload exposes it. The component autocoder now generates a public Validate_Parameter_Defaults subprogram that dispatches to Validate_Parameters with the working-copy values and fails an assertion on an invalid result, and the assembly autocoder calls it for every parameterized instance during the Set_Id_Bases phase. Running before components are connected or initialized is deliberate: it makes init-order independence part of the Validate_Parameters contract, so an override that strays into invoking a connector trips the unconnected-connector assertion at first startup rather than misbehaving later; the generated comment above Validate_Parameters now states that contract. Dying on an invalid default is likewise intentional - a default the component itself rejects is a defective build, and the assertion surfaces it on the first unit-test or simulation run instead of letting it reach flight.

### DIFF
--- a/gen/templates/assembly/name.adb
+++ b/gen/templates/assembly/name.adb
@@ -50,6 +50,13 @@ package body {{ name }} is
 {% for component in component_kind_dict["data_dependencies"] %}
       {{ component.instance_name }}.Map_Data_Dependencies{% if component.map_data_dependencies.parameter_call_string() %} ({{ component.map_data_dependencies.parameter_call_string() }}){% endif %};
 {% endfor %}
+
+      -----------------------------------
+      -- Default Parameter Validation:
+      -----------------------------------
+{% for component in component_kind_dict["parameters"] %}
+      {{ component.instance_name }}.Assert_Valid_Parameters;
+{% endfor %}
    end Set_Id_Bases;
 
 {% endif %}
--- a/gen/templates/component/component-name.adb
+++ b/gen/templates/component/component-name.adb
@@ -271,6 +271,27 @@ package body Component.{{ name }} is
    end Map_Data_Dependencies;
 
 {% endif %}
+{% if parameters %}
+   -----------------------------------------------------------------------
+   -- Assert that the component's default parameter values are valid:
+   -----------------------------------------------------------------------
+   not overriding procedure Assert_Valid_Parameters (Self : in out Base_Instance) is
+      use Parameter_Validation_Status;
+   begin
+      -- An assertion failure here means the default parameter values of {{ name }}
+      -- failed validation. The assertion carries no message string to keep it out of
+      -- the binary; the failure is traceable through this subprogram's symbol and the
+      -- assertion's file and line.
+      pragma Assert (
+         Base_Instance'Class (Self).Validate_Parameters (
+{% for par in parameters %}
+            {{ par.name }} => Self.{{ par.name }}{{ "," if not loop.last }}
+{% endfor %}
+         ) = Valid
+      );
+   end Assert_Valid_Parameters;
+
+{% endif %}
 {% if connectors.requires_queue() %}
    ---------------------------------------------------------------
    -- Visible private dispatching procedures for component queue:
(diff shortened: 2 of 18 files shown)


gen/templates/component/component-name.adb
```diff
@@ -270,6 +270,25 @@ package body Component.{{ name }} is
...
+   not overriding procedure Validate_Parameter_Defaults (Self : in out Base_Instance) is
+      use Parameter_Validation_Status;
+   begin
+      pragma Assert (
+         Base_Instance'Class (Self).Validate_Parameters (
+{% for par in parameters %}
+            {{ par.name }} => Self.{{ par.name }}{{ "," if not loop.last }}
+{% endfor %}
+         ) = Valid,
+         "{{ name }} default parameter validation failed."
```
dinkelk: so we try not to put strings in our asserts in flight code and use comment instead. this is just to keep the strings out of the binary, which can actually bloat significantly when you write nice messages for all asserts.

gen/templates/component/component-name.adb
```diff
@@ -270,6 +270,25 @@ package body Component.{{ name }} is
...
+      );
+      null;
```
dinkelk: delete.

gen/templates/component/component-name.adb
```diff
@@ -270,6 +270,25 @@ package body Component.{{ name }} is
       );
    end Map_Data_Dependencies;
 
+{% endif %}
+{% if parameters %}
+   -----------------------------------------------------------------------
+   -- Validate the default values of the component's parameters:
+   -----------------------------------------------------------------------
+   not overriding procedure Validate_Parameter_Defaults (Self : in out Base_Instance) is
```
dinkelk: maybe rename Assert_Valid_Parameters

gen/templates/assembly/name.adb
dinkelk, line 30: We need this new feature to run in unit tests as well. Make modification to tester component implementation template (component-name-implementation-tester.adb) to add a call to this new parameter assertion method as the last line of Init_Base. We will need to update existing components that have parameters to get this update to their templates across Adamant and project repos.

dinkelk (changes requested):

gen/templates/assembly/name.adb
cbrisebois-lasp: exploring this

gen/templates/component/component-name.adb
cbrisebois-lasp: done

gen/templates/component/component-name.adb
cbrisebois-lasp: removing

gen/templates/component/component-name.adb
cbrisebois-lasp: deleted

gen/templates/assembly/name.adb
cbrisebois-lasp: okay, I came up with a bit of a hybrid solution. So as far as I can tell the only components that don't have there bases set up in the auto code are generics. So for every regular component I just added the call to the auto code, and for generics I added a place for the user to implement in. The only parameterized component that fell into this category was the limiter, so it got at update, but outside of that no other changes needed.

Not sure if this is what you had in mind. But it seems pretty tidy?

gen/templates/tests/name-implementation.adb
```diff
@@ -23,6 +23,13 @@ package body {{ name }}.Implementation is
...
 
+{% if component.parameters %}
+      -- Assert that the component's compiled-in default parameter values are valid.
+      -- The generated test base performs this check for non-generic components; a
+      -- generic component's tester only exists here, so the call is made here.
+      Self.Tester.Component_Instance.Assert_Valid_Parameters;
+
+{% endif %}
 {% endif %}
 {% if component.connectors.invoker() or component.connectors.of_kind("recv_async") or component.events %}
```
dinkelk, line 34: The alternative would be to put Self.Tester.Component_Instance.Assert_Valid_Parameters; inside of component-name.adb in Init_Base and fix this line to say `or component.parameters`

PR #208, opened 2026-08-21
Cache computed type ranges by Ada type name within a redo session

dinkelk:
## Problem

Producing a `*.type_ranges.yaml` file requires generating, compiling, linking, and running a dedicated Ada program, once per packed type model. These programs exist to extract ground truth from the compiler (`'First`/`'Last` bounds and enumeration literals via `'Image`/`'Enum_Rep`), but most packed type models draw their fields from the same small set of underlying Ada types (`Natural`, `Interfaces.Unsigned_16`, `Basic_Types.Byte`, and so on). On a clean build, nearly every one of these programs relinks an executable to recompute answers that another model's program already produced. Models whose fields are all packed types or modeled enums pay the full compile and link cost just to print an empty list.

## Approach

A session-scoped cache database (`type_ranges_cache_database`, following the `model_cache_database` pattern) maps Ada type names to their computed ranges. It is consulted in `packed_type.load_type_ranges()`:

- The stand-in type ranges model handed to `set_type_ranges()` resolves each `get_type_by_name()` call against the cache as it is made. If every name is served, the model's type ranges program is never built or run. Models that ask for nothing (all fields packed types or modeled enums) skip the program without a single cache lookup.
- A name the cache cannot serve raises an internal miss exception, and `load_type_ranges()` falls back to the existing path: build the yaml via the program, load it, then store every cacheable type from the result for the rest of the session.

Resolving lazily rather than predicting which fields need ranges is what keeps the feature contained. The questions the cache answers are precisely the questions `set_type_ranges()` asks, so no filter has to be duplicated and no packed type subclass needs to know the cache exists: `record.py`, `array.py` and `type.py` are untouched. The `build_type_ranges_yaml` rule is untouched too, so explicitly redoing a `*.type_ranges.yaml` target behaves exactly as before. The one piece of shared knowledge, the native platform directory the programs are built for, is now a single `target.get_native_build_for()` helper used by both the rule that locates a program and the cache that keys its entries.

## Performance

Measured on a cleaned `src/types` subtree (103 packed type models), both arms run back to back from identical clean states:

| | Wall time | Type ranges programs built |
|---|---|---|
| `DISABLE_TYPE_RANGES_CACHE=1` | 9m37s | 103 of 103 |
| cache enabled | 3m22s | 35 of 103 |

2.9x faster here, with 66% of the programs never built, and identical range output from both arms.

### DIFF
--- a/gen/models/packed_type.py
+++ b/gen/models/packed_type.py
@@ -3,8 +3,64 @@ from util import ada
 from util import redo_arg
 from util import model_loader
 from util import redo
+from collections import OrderedDict
 import os
 import abc
+
+
+class _type_ranges_cache_miss(Exception):
+    """
+    Raised internally when a type range cannot be served from the
+    session cache. It tells load_type_ranges() to fall back to building
+    and running this model's own type ranges program.
+    """
+    pass
+
+
+class _cached_type_ranges(object):
+    """
+    A stand-in for a type_ranges model that resolves lookups from the
+    session type ranges cache instead of from a *.type_ranges.yaml file.
+
+    This is what lets the cache work without any packed type subclass
+    knowing it exists. Every set_type_ranges() implementation already
+    asks a type ranges model for exactly the types it needs, one name at
+    a time, so serving those same calls from the cache requires no
+    duplicate knowledge of which fields need ranges. A name that cannot
+    be served raises _type_ranges_cache_miss.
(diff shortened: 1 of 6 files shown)


Jbsco (approved):
This is great, noting considerable time reductions in my environment as well.

PR #210, opened 2026-08-25
Convert the binary tree package to SPARK silver

dinkelk:
## Summary

Converts `src/data_structures/binary_tree/` to SPARK and proves it at the **silver** level (absence of runtime errors) with GNATprove 15.1.0 at `--level=2`.

## What changed

- **API unchanged.** Every subprogram keeps its name, profile and behavior. All contracts are proof only: an `Assertion_Policy` at the top of the package ignores them at runtime, so the generated code is identical to the unconverted package for every instantiation. The defensive `pragma Assert` statements in `Add` and `Search` stay live and are now proved.
- **Every contract and loop invariant carries a one sentence plain English comment** directly above it saying what it checks.
- **One ghost predicate, `Is_Valid`**: storage allocated, indexed from 1, within `Max_Tree_Size`, and `Size <= Tree'Last`. `Init` establishes it, everything else preserves it. With it, the size assertion at the top of `Search` and every index inside the binary search are proved.
- **`Init` and `Destroy`** are the only `SPARK_Mode => Off` bodies (heap allocation and conditional deallocation are outside the ownership model). Their postconditions are documented in the body.
- **`Get` and `Set`** gain the explicit precondition `Element_Index <= Get_Size (Self)`, which is the obligation callers already meet by taking indices from `Search` or `Get_First_Index .. Get_Last_Index`.
- **`Max_Tree_Size = Positive'Last - 1`**: the binary search steps `Mid_Index + 1`, which genuinely overflows when the capacity is `Positive'Last`. Trees above the bound are outside the verified domain and behave exactly as before.
- `Get_Size` and `Get_Capacity` become expression functions in the private part so the proof can see through them. Functions with `in out`/`out` parameters carry the `Side_Effects` aspect SPARK requires.
- `all.prove.yaml` selects silver, level 2. `binary_tree_prover.ads` instantiates the generic so GNATprove has an instance to analyze (generics are only analyzed at instantiation points). Nothing references it and it adds no code to any build.

## Proof results

```
Run-time Checks       55   proved
Assertions             9   proved
Functional Contracts  16   proved
Initialization         3   flow
Termination            9   8 flow, 1 proved
Total                 92   0 unproved, 0 justified, 0 pragma Assume
```

### DIFF
--- a/src/data_structures/binary_tree/all.prove.yaml
+++ b/src/data_structures/binary_tree/all.prove.yaml
@@ -1,0 +1,4 @@
+---
+description: GNATprove configuration for the binary tree package. The target is silver, absence of runtime errors, including proof of the defensive assertions.
+level: 2
+mode: "silver"
--- a/src/data_structures/binary_tree/binary_tree.adb
+++ b/src/data_structures/binary_tree/binary_tree.adb
@@ -1,16 +1,23 @@
 with Safe_Deallocator;
 
-package body Binary_Tree is
+package body Binary_Tree with SPARK_Mode => On is
    ----------------------------------
    -- Public sub programs:
    ----------------------------------
 
-   procedure Init (Self : in out Instance; Maximum_Size : in Positive) is
+   -- The body is not analyzed by SPARK since it performs the heap allocation
+   -- that owns the element storage for the rest of the tree's life. The
+   -- postcondition holds because the precondition provides an empty tree and
+   -- the allocation below is indexed from 1 with the requested positive size.
+   procedure Init (Self : in out Instance; Maximum_Size : in Positive) with SPARK_Mode => Off is
    begin
       Self.Tree := new Element_Array (Positive'First .. Positive'First + Maximum_Size - 1);
    end Init;
 
-   procedure Destroy (Self : in out Instance) is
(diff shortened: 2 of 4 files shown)


src/data_structures/binary_tree/all.prove.yaml
```diff
@@ -0,0 +1,4 @@
+---
+description: GNATprove configuration for the binary tree package. The target is silver, absence of runtime errors, including proof of the defensive assertions.
+level: 3
```
Jbsco: This also proves at effort level 2 (also level 0), same result and duration as 3, locally. I don't think there's any penalty or disadvantage, just noting we seem to have a default of 2 so far, and noticed this was higher but uncertain why.
Maybe worth discussion... I believe failure to prove a condition due to timeout is surfaced in logs, if a future change increased condition complexity and the prover failed at level 2 (or 0), triage would likely be an easy bump up, maybe we gain some insight in those cases.

Jbsco (approved):

src/data_structures/binary_tree/all.prove.yaml
dinkelk: will turn down to 2.

PR #211, opened 2026-08-25
Assert valid parameters

cbrisebois-lasp:


dinkelk:
Hm. Not sure I love it. I think I confused you.

The current job of the component init base is to initialize heap items, full stop. Adding parameter validation there does not seem to fit. I like it much better with the IDed entity initialization in set_id_bases. Since parameters are IDed entities this seems synergistic.

I originally asked you to add the call to validate parameters in the TESTER's autocoded init_base, as a short cut to get default parameter validation during unit test for free. I never intended for you to put validation in the component's init_base. I am sorry to confuse there. Let me know what you think. I think the previous design was actually perfect, we just didn't have a hook for the unit test, that was the only thing missing.

cbrisebois-lasp:
@dinkelk not sure why I can't directly reply to your comment. Anyways, yes! This is exactly what I was thinking. I thought the encapsulation was weird. I kept the original implementation around just incase because this didn't feel right. 

Okay so revert to previous implementation, but change the testers init_base to call the assert_validate_parameters?

PR #212, opened 2026-08-26
Convert the protected variables package to SPARK silver

dinkelk:
## Summary

Converts `src/data_structures/protected_variables/` to SPARK and proves it at the **silver** level (absence of runtime errors) with GNATprove 15.1.0 at `--level=3`. First SPARK proof of protected types in the framework. Single commit.

```
Run-time Checks       10   proved
Initialization         3   flow
Termination            9   flow
Total                 22   0 unproved, 0 justified, 0 pragma Assume
```

## What changed

- **Four of five generics prove with no contracts at all**, just `SPARK_Mode => On`: `Generic_Variable`, `Generic_Protected_Counter` (modular, wraps by definition), `Generic_Staged_Variable`, `Generic_Protected_Periodic_Counter` (modular; its `mod Period` was already guarded by `Period > 0`). No assertion policy pragma is needed since there are no contracts.
- **`Generic_Protected_Counter_Decrement` becomes a countdown that saturates at zero.** This is the one real finding: `Count := @ - To_Subtract` on a signed `T` could leave the range and raise `Constraint_Error` inside the protected action. A precondition is not an option here: Ada evaluates a protected operation's `Pre` *outside* the lock, so it may not depend on the object's state (GNAT: `internal call cannot appear in precondition of protected operation`), which is also the reason a caller-side check would be a time-of-check/time-of-use race. The counter's only user treats it as a countdown of remaining samples with zero meaning "expired", so the type now says so:
  - decrementing by more than the current count leaves it at zero; no operation raises;
  - `pragma Compile_Time_Error` requires `T'First = 0` and `T'Last >= 1`, so zero is the floor and the default decrement of 1 is representable (verified to reject an `Integer` instantiation with the message);
  - the spec comments describe the saturation for callers, and the unit test exercises saturation from zero, from a small count, and via `Decrement_Count_And_Return_Previous`;
  - two comments that wrongly said "Any modular type" / "Increment the count" on this signed decrementing generic are corrected.
- **pid_controller** (the sole instantiator in this repo and in the downstream projects I can see) moves its diagnostic countdown from `Integer` to `Natural`. Behavior unchanged: it already guards with `if Diag_Count > 0`. Its 9 unit tests pass.
- **`Linux_Prove` gets its own `linux_prove.gpr`**: `linux_debug.gpr` plus the existing `sequential_elaboration.adc`. GNATprove refuses tasking constructs without `pragma Partition_Elaboration_Policy (Sequential)` (SPARK RM 9(2)); the bareboard flight targets already compile with exactly this pair (`ravenscar.adc` + `sequential_elaboration.adc`), so the prove configuration now matches the flight one more closely. `Linux_Debug`/`Linux_Test` are untouched. `stopwatch` and the doc example still prove under the new project.
- `all.prove.yaml` (silver, level 3) and `protected_variables_prover.ads`, which instantiates every generic (the countdown twice, with `Natural` and a narrow `0 .. 7` range) so GNATprove has instances to analyze. Referenced by nothing, adds no code.

## Note for downstream projects

A project that ships its own `Linux_Prove` target (overriding the framework's) needs the same two-line GPR addition to prove protected types locally; `prove_all` in this repository's CI uses the framework target and needs nothing.

## Verification

- `redo prove` in `src/data_structures/protected_variables`: clean at silver level 3.
- `redo test` in `src/data_structures/protected_variables/test`: 5/5 pass (decrement test extended with saturation cases).
- `redo test` in `src/components/pid_controller/test`: 9/9 pass.
- `redo style` on both directories: clean.
- `redo all` for command_router, cpu_monitor, queue_monitor, connector_queuer (instantiators of the other generics): builds.
- `Compile_Time_Error` confirmed to reject `Generic_Protected_Counter_Decrement (Integer)`.

### DIFF
--- a/redo/targets/gpr/linux_prove.gpr
+++ b/redo/targets/gpr/linux_prove.gpr
@@ -1,0 +1,33 @@
+project linux_prove extends all "a_linux_debug_base.gpr" is
+
+   -----------------------------------------------
+   -- These lines of code must be included at the
+   -- top of every Adamant based .gpr file. They
+   -- are used to connect the Adamant build system
+   -- to GPRBuild.
+   -----------------------------------------------
+   for Source_Dirs use a_adamant.SOURCE_DIRS;
+   for Excluded_Source_Files use a_adamant.EXCLUDED_SOURCE_FILES;
+   for Object_Dir use a_adamant.OBJECT_DIR;
+   for Exec_Dir use a_adamant.EXEC_DIR;
+
+   --
+   -- This GPR file is the same as linux_debug.gpr but also enforces the
+   -- sequential partition elaboration policy, as the bareboard targets do.
+   -- GNATprove requires it to analyze tasking constructs such as protected
+   -- types (SPARK RM 9(2)), and it matches the configuration the flight
+   -- software actually runs under.
+   --
+   package Compiler is
+      -- Add preprocessor definitions and configuration pragma switches:
+      for Switches ("Ada") use a_linux_debug_base.Compiler'Switches ("Ada") &
+         -- add ravenscar restriction
+         ("-gnatec=" & a_adamant.ADAMANT_DIR & "/redo/targets/configuration_pragmas/ravenscar.adc") &
+         -- add sequential elaboration policy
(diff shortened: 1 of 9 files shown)


dinkelk:
Moving to the dinkelk fork for review first; will re-open here once reviewed.

PR #213, opened 2026-08-26
Convert the protected variables package to SPARK silver

dinkelk:
## Summary

Converts `src/data_structures/protected_variables/` to SPARK and proves it at the **silver** level (absence of runtime errors) with GNATprove 15.1.0 at `--level=2`. First SPARK proof of protected types in the framework. Two commits: the `Linux_Prove` project file that GNATprove needs for tasking, then the package conversion.

## What changed

- **`Linux_Prove` gets its own `linux_prove.gpr`**: `linux_debug.gpr` which includes `sequential_elaboration.adc`. GNATprove requires `pragma Partition_Elaboration_Policy (Sequential)` to analyze tasking constructs (SPARK RM 9(2)). The bareboard targets already compile with this pair of pragmas. Debug and test builds are untouched; existing SPARK packages still prove.
- **`Generic_Protected_Counter_Decrement` is now a countdown that saturates at zero.** Its subtraction could leave the range of `T` and raise `Constraint_Error` inside the protected action. Decrementing past zero now leaves the count at zero, nothing raises.
- **pid_controller**, the only instantiator, moves from `Integer` to `Natural` and uses `Decrement_Count_And_Return_Previous`, replacing a read followed by a separately guarded decrement with one protected action. Control flow is otherwise unchanged.

### DIFF
--- a/redo/targets/gpr/linux_prove.gpr
+++ b/redo/targets/gpr/linux_prove.gpr
@@ -1,0 +1,33 @@
+project linux_prove extends all "a_linux_debug_base.gpr" is
+
+   -----------------------------------------------
+   -- These lines of code must be included at the
+   -- top of every Adamant based .gpr file. They
+   -- are used to connect the Adamant build system
+   -- to GPRBuild.
+   -----------------------------------------------
+   for Source_Dirs use a_adamant.SOURCE_DIRS;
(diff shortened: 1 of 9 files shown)


Jbsco (approved):
Clean!

PR #214, opened 2026-08-26
Assert valid parameters

cbrisebois-lasp:
### Assert component default parameter values via Validate_Parameters

A component's Validate_Parameters function today runs only when a parameter operation arrives from the ground, so the default parameter values compiled in from the component model are never checked against any custom validation the developer implements there. An invalid default flies silently until the first parameter update. This PR closes that gap by asserting the defaults through Validate_Parameters both at assembly startup and in component unit tests.

**What's generated**

Components with parameters gain one new base-package primitive:

- Assert_Valid_Parameter_Defaults — asserts on a dispatching Validate_Parameters call against the component's working parameter copies, which hold the model defaults until the first parameter update. It lives in the base package because the working copies are traceable through the subprogram symbol and the assert's file and line.

**Where it's called**

- Startup: at the end of Set_Id_Bases — the one initialization step every parameterized component has (parameters always contribute Parameter_Id_Base), unlike Init_Base, which only exists for components needing heap setup.
- Unit tests: for a regular component, the generated tests base calls it in Set_Up right after allocating the tester — the tests base regenerates on every build, so existing tests inherit the check without touching their committed testers. For a generic component the tester is instantiated inside the developer's test implementation where the tests base cannot reach, so the call is emitted into the tester's Init_Base template instead and reaches existing generic tests as their testers are regenerated (the limiter is the only generic parameterized component in the repo, and is updated in this PR). Both hooks run before Set_Up_Test, while the working copies are still pristine defaults.

**Design contract**

Because the check runs before the component is connected or initialized, Validate_Parameters must be a pure function of its arguments — no dependence on Init state, no connector invocations. This was already the intended shape of the function; the calls make it load-bearing, and it is now documented at the override point in both the base spec and the implementation template.

**Review notes**

The four commits layer for review: (1) the primitive alone, (2) the startup integration, (3) the test integration, and (4) the patch to the limiter component. Generated doc comments in each commit describe only the behavior that exists at that point.

**Verification**

- Unit tests pass for every parameterized component in the repo: limiter (generic path, tester Init_Base hook), pid_controller (tests-base Set_Up hook), and the parameters component's test and test_grouped suites, whose testers call Set_Id_Bases directly and so exercise the startup assert at runtime.
- Negative proof: with a component's model default deliberately edited so its Validate_Parameters rejects it, the regenerated unit test fails during Set_Up with ADA.ASSERTIONS.ASSERTION_ERROR pointing at the generated assert; restoring the default returns the suite to green.
- A full downstream assembly build compiles cleanly with the regenerated autocoder output.

### DIFF
--- a/gen/templates/component/component-name.adb
+++ b/gen/templates/component/component-name.adb
@@ -241,6 +241,9 @@ package body Component.{{ name }} is
       pragma Assert (Natural (Parameter_Id_Base) + {{ (parameters|length) - 1 }} <= Natural (Parameter_Types.Parameter_Id'Last));
 {% endif %}
       Self.Parameter_Id_Base := Parameter_Id_Base;
+      -- Check the component's compiled-in default parameter values against the
+      -- component's parameter validation:
+      Self.Assert_Valid_Parameter_Defaults;
 {% endif %}
    end Set_Id_Bases;
 
--- a/gen/templates/component/component-name.adb
+++ b/gen/templates/component/component-name.adb
@@ -1054,6 +1054,16 @@ package body Component.{{ name }} is
    -- Parameter handling primitives:
    -------------------------------------------
 
+   not overriding procedure Assert_Valid_Parameter_Defaults (Self : in out Base_Instance) is
+      use Parameter_Validation_Status;
+   begin
+      pragma Assert (Base_Instance'Class (Self).Validate_Parameters (
+{% for par in parameters %}
+         {{ par.name }} => Self.{{ par.name }}{{ "," if not loop.last }}
+{% endfor %}
+      ) = Valid);
+   end Assert_Valid_Parameter_Defaults;
+
    not overriding procedure Update_Parameters (Self : in out Base_Instance) is
    begin
       -- If parameters have finished being staged, then we need to update
--- a/gen/templates/component/component-name-implementation.ads
+++ b/gen/templates/component/component-name-implementation.ads
@@ -175,7 +175,10 @@ private
    -- overridden if something special needs to happen to further validate a parameter. Examples of this might be validation of
    -- certain parameters beyond individual type ranges, or performing other special functionality that only needs to be
    -- performed after parameters have been validated. Note that range checking is performed during staging, and does not need
-   -- to be implemented here.
+   -- to be implemented here. This function is also called through Assert_Valid_Parameter_Defaults from Set_Id_Bases, before
+   -- the component is connected or initialized, to check the compiled-in default parameter values. The implementation must
+   -- therefore be a pure function of the passed-in parameter values, with no dependence on Init state and no connector
+   -- invocations.
    overriding function Validate_Parameters (
       Self : in out Instance;
 {% for par in parameters %}
--- a/gen/templates/component/component-name.ads
+++ b/gen/templates/component/component-name.ads
@@ -274,7 +274,10 @@ package Component.{{ name }} is
    -- overridden if something special needs to happen to further validate a parameter. Examples of this might be validation of
    -- certain parameters beyond individual type ranges, or performing other special functionality that only needs to be
    -- performed after parameters have been validated. Note that range checking is performed during staging, and does not need
-   -- to be implemented here.
+   -- to be implemented here. This function is also called through Assert_Valid_Parameter_Defaults from Set_Id_Bases, before
+   -- the component is connected or initialized, to check the compiled-in default parameter values. The implementation must
+   -- therefore be a pure function of the passed-in parameter values, with no dependence on Init state and no connector
+   -- invocations.
    not overriding function Validate_Parameters (
       Self : in out Base_Instance;
 {% for par in parameters %}
@@ -285,6 +288,8 @@ package Component.{{ name }} is
    -- This procedure asserts that the component's working parameter values, which
    -- hold the defaults declared in the component model until the first parameter
    -- update, are valid according to the component's Validate_Parameters function.
+   -- Set_Id_Bases calls it at startup, checking the model defaults against any
+   -- custom validation.
    not overriding procedure Assert_Valid_Parameter_Defaults (Self : in out Base_Instance);
 
 {% endif %}
--- a/gen/templates/component/component-name.ads
+++ b/gen/templates/component/component-name.ads
@@ -282,6 +282,11 @@ package Component.{{ name }} is
 {% endfor %}
    ) return Parameter_Validation_Status.E is abstract;
 
+   -- This procedure asserts that the component's working parameter values, which
+   -- hold the defaults declared in the component model until the first parameter
+   -- update, are valid according to the component's Validate_Parameters function.
+   not overriding procedure Assert_Valid_Parameter_Defaults (Self : in out Base_Instance);
+
 {% endif %}
 {% if data_dependencies %}
    -----------------------------------------------


gen/templates/component/component-name.adb
dinkelk, line 202: This is subtle. But can you look at what triggers this to have content in the component.py model. We need to mere existence of a parameters model to trigger this, even if a component doesn't have a Parameter_Update invoker connector. Weird case, but I want to make sure it is covered. 

Currently the check is
```
365         if self.set_id_bases_parameters:
```

to which we might need to add `or self.parameters` or maybe CLAUDE has a more accurate solution here.

dinkelk (changes requested):

gen/templates/component/component-name.adb
cbrisebois-lasp: Confirmed. Here's the path:

1. gen/models/component.py:134 — set_id_bases_parameters starts empty during component load.
2. gen/models/component.py:331 — every submodel attached to the component gets submodel.set_component(self) called.
3. gen/models/parameters.py:44-51 — the parameters submodel's set_component unconditionally appends parameter_Id_Base to component.set_id_bases_parameters as its first statement. No branch, no way to skip it.
4. gen/models/component.py:365-375 — runs after the submodel loop; since the list is non-empty, self.set_id_bases is set to a subprogram object (truthy). The None branch is unreachable for a parameterized component.

No need for `or self.parameters` there

gen/templates/tests/name.adb
```diff
@@ -74,6 +74,11 @@ package body {{ name }} is
       Self.Tester := Tester_Alloc.Allocate;
       -- Link the log access type to the logger in the reciprocal
       Self.Tester.Set_Logger (Self.Logger'Unchecked_Access);
+{% if component.parameters %}
+      -- Check the component's compiled-in default parameter values against the
+      -- component's parameter validation:
+      Self.Tester.Component_Instance.Assert_Valid_Parameter_Defaults;
```
Jbsco, line 80: This is also called at [component-name-implementation-tester.adb:32](https://github.com/lasp/adamant/pull/214/changes#diff-92cb3be49d890b67109cddd7952dd8b1c758a496cb46b5455786a50e3111c42fR32), so would it run twice for every component with parameters? Might be able to drop the tester call, the one in `name.adb` also covers tests that never call `Init_Base`.

gen/templates/component/component-name.adb
```diff
@@ -1054,6 +1057,16 @@ package body Component.{{ name }} is
...
    -------------------------------------------
 
+   not overriding procedure Assert_Valid_Parameter_Defaults (Self : in out Base_Instance) is
+      use Parameter_Validation_Status;
+   begin
+      pragma Assert (Base_Instance'Class (Self).Validate_Parameters (
+{% for par in parameters %}
+         {{ par.name }} => Self.{{ par.name }}{{ "," if not loop.last }}
+{% endfor %}
+      ) = Valid);
```
Jbsco, line 1067: Maybe superfluous, but could add a message with the component type.

Jbsco (changes requested):

gen/templates/tests/name.adb
cbrisebois-lasp: good catch, fixing!

gen/templates/component/component-name.adb
cbrisebois-lasp: I think this would be a useful feature. Adding!

gen/templates/component/component-name.adb
```diff
@@ -1054,6 +1057,16 @@ package body Component.{{ name }} is
...
    -------------------------------------------
 
+   not overriding procedure Assert_Valid_Parameter_Defaults (Self : in out Base_Instance) is
+      use Parameter_Validation_Status;
+   begin
+      pragma Assert (Base_Instance'Class (Self).Validate_Parameters (
+{% for par in parameters %}
+         {{ par.name }} => Self.{{ par.name }}{{ "," if not loop.last }}
+{% endfor %}
+      ) = Valid, "{{ name }} parameter defaults failed Validate_Parameters (Parameter_Id_Base" & Parameter_Types.Parameter_Id'Image (Self.Parameter_Id_Base) & ").");
```
dinkelk: no assertion message.

gen/templates/tests/component-name-implementation-tester.adb
```diff
@@ -25,6 +25,14 @@ package body Component.{{ name }}.Implementation.Tester is
       -- Initialize component heap:
       Self.Component_Instance.Init_Base{% if init_base.parameters %} ({% for n in init_base.parameter_names %}{{ n }} => {{ n }}{{ ", " if not loop.last }}{% endfor %}){% endif %};
 
+{% endif %}
+{% if parameters and generic %}
+      -- Check the component's compiled-in default parameter values against the
+      -- component's parameter validation. Generic components are checked here
+      -- because the base test Set_Up cannot allocate their tester; non-generic
+      -- components are checked in the base test Set_Up instead.
+      Self.Component_Instance.Assert_Valid_Parameter_Defaults;
```
dinkelk: I changed my mind on this again. Let's talk. After working through very similar problem here:

https://github.com/lasp/adamant/pull/220/changes#diff-868ba1ecc5ffc170a04780111878e2fda625e88d13a8c94cb7de48fe162a2bceR31

I think adding into test implementation template is better instead of tester. This works, but is not symetric, and thus confusing. If the test does this for normal components, then the test-implementation should probably be doing this for generics.

dinkelk (changes requested):

gen/templates/tests/name-implementation.adb
```diff
@@ -34,7 +34,7 @@ package body {{ name }}.Implementation is
       Self.Tester.Set_Logger (Self.Logger'Unchecked_Access);
```
dinkelk, line 35: Add
```
{% if component.parameters %}
      -- Check the component's compiled-in default parameter values against the
      -- component's parameter validation:
      Self.Tester.Component_Instance.Assert_Valid_Parameter_Defaults; 
{% endif %}
```

gen/templates/tests/component-name-implementation-tester.adb
```diff
@@ -25,6 +25,14 @@ package body Component.{{ name }}.Implementation.Tester is
       -- Initialize component heap:
       Self.Component_Instance.Init_Base{% if init_base.parameters %} ({% for n in init_base.parameter_names %}{{ n }} => {{ n }}{{ ", " if not loop.last }}{% endfor %}){% endif %};
 
+{% endif %}
+{% if parameters and generic %}
```
dinkelk: This logic is just bent pipe for generic param validation. We are doing this in name-implementation.adb instead, so remove this logic and all the `or parameters` that feed it.

gen/templates/tests/name-implementation.adb
dinkelk, line 29: This is already gaurded, so if you add the param validation for generic specifically you are good

gen/templates/tests/name-implementation.adb
dinkelk: already guarded as generic pattern only

gen/templates/component/component-name.adb
cbrisebois-lasp: removing

dinkelk (approved):

Jbsco (approved):

PR #215, opened 2026-08-27
Convert the CRC, checksum and parity packages to SPARK silver

dinkelk:
## Summary

Converts `src/util/crc/` (`Crc_16`, `Checksum_16`, `Xor_8`) to SPARK and proves them at the **silver** level (absence of runtime errors) with GNATprove 15.1.0 at `--level=2`. Single commit.

```
Run-time Checks       12   proved
Total                 16   0 unproved, 0 justified, 0 pragma Assume
```

## What changed

- **`SPARK_Mode => On` on all three packages.** They prove as written: loops over a byte array with table lookups and modular arithmetic, so no contracts, loop invariants or assertion policy are needed.
- **`Compute_Crc_16 (Byte_Array_Pointer.Instance)` body is `SPARK_Mode => Off`.** It overlays the pointed-to memory by address so a CRC can be taken at address zero, which is outside the SPARK memory model. It delegates to the proved array version.
- `all.prove.yaml` (silver, level 2).

### DIFF
--- a/src/util/crc/all.prove.yaml
+++ b/src/util/crc/all.prove.yaml
@@ -1,0 +1,4 @@
+---
+description: GNATprove configuration for the CRC, checksum and parity packages. The target is silver, absence of runtime errors.
+level: 2
+mode: "silver"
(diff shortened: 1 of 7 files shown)


Jbsco (approved):
Straightforward, nice.

PR #216, opened 2026-08-27
Convert the byte array utilities to SPARK silver

dinkelk:
## Summary

Converts `src/util/byte_array_util/` to SPARK and proves it at the **silver** level (absence of runtime errors) with GNATprove 15.1.0 at `--level=2`. Single commit.

```
Run-time Checks      109   proved
Assertions             4   proved
Functional Contracts   2   proved
Total                120   0 unproved, 0 justified, 0 pragma Assume
```

## What changed

- **`SPARK_Mode => On` and `Side_Effects`** on the four functions with `out`/`in out` parameters. The bit extraction and insertion loops prove as written; no loop invariants were needed. Contracts are proof-only (`Assertion_Policy` as in binary_tree): the two postconditions on the copy functions, previously live, are now guaranteed by the proof rather than a runtime check. The defensive `pragma Assert`s stay live and are proved.
- **Overflow in the size checks fixed.** `Offset + Size > Src'Length * 8` overflows `Integer` for an offset near `Integer'Last` or an array over 256 MB, raising `Constraint_Error` where `Error` should be returned. The comparison is now done in 64 bits, and the index arithmetic splits the offset into whole bytes and remaining bits so it cannot overflow either. Results are identical for every input the old code handled.
- **`Signed_Length` removed.** It could overflow for an empty array with far apart bounds, and the copy postconditions written with it were false for such arrays (copy returns 0, `Signed_Length` is negative). Nothing outside the package used it. The postconditions and internal assertions now say what they meant: no more bytes are copied than either array holds (`'Length`).
- `all.prove.yaml` (silver, level 2).

### DIFF
--- a/src/util/byte_array_util/byte_array_util.adb
+++ b/src/util/byte_array_util/byte_array_util.adb
@@ -1,6 +1,18 @@
 with Interfaces; use Interfaces;
 
-package body Byte_Array_Util is
+package body Byte_Array_Util with SPARK_Mode => On is
+
+   -- See the note in the package specification. All contracts are for proof
+   -- only and are disabled at runtime:
+   pragma Assertion_Policy
+      (Pre => Ignore,
+       Post => Ignore,
+       Contract_Cases => Ignore,
+       Ghost => Ignore,
+       Loop_Invariant => Ignore,
+       Loop_Variant => Ignore,
+       Assert_And_Cut => Ignore,
+       Assume => Ignore);
 
    function Safe_Right_Copy (Dest : in out Byte_Array; Src : in Byte_Array) return Natural is
    begin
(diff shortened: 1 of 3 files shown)


src/util/byte_array_util/byte_array_util.adb
```diff
@@ -78,9 +90,10 @@ package body Byte_Array_Util is
       Value := [0, 0, 0, 0];
 
       -- Validate offset and size. Size must be <= 32 bits. The size + offset must not overflow
-      -- the size of the byte array. Size must be greater than zero.
+      -- the size of the byte array. Size must be greater than zero. The comparison is done in
+      -- 64 bits so that it cannot itself overflow for any offset or array length.
       if Size > Poly_32_Type'Object_Size or else
-          Offset + Size > Src'Length * Byte'Object_Size
+          Long_Long_Integer (Offset) + Long_Long_Integer (Size) > Long_Long_Integer (Src'Length) * Long_Long_Integer (Byte'Object_Size)
```
Jbsco, line 96: Good catch!

Jbsco (approved):
These are great.
I almost suggested adding a test that would have caught the potential overflow when calculating in `Natural`, but now it's SPARK, so... ✨

PR #218, opened 2026-08-27
Convert the variable database package to SPARK silver

dinkelk:
Convert the variable database package to SPARK silver

Prove the generic Variable_Database package free of runtime errors with
GNATprove at the silver level. This is the store beneath the product
database component.

The public API is unchanged and the contracts are proof only: an
Assertion_Policy at the top of the package ignores them at runtime, so
the generated code is identical to the unconverted package. The proof
rests on the ghost predicate Is_Valid, stating that the table has been
allocated with bounds within Id_Type, and on ghost accessors for the
range of Ids held, which every operation is proved to preserve. The
status contracts state that an operation fails with Id_Out_Of_Range
exactly when the Id is outside that range, which is what the defensive
assertion in Clear_Override_All relies on.

The variable serializer is instantiated directly, as before. Its
declarations are in SPARK and its body is not analyzed, so the store and
load of a value are proved against its declarations and treated as
trusted boundary operations. Init and Destroy are excluded from
analysis, for the heap allocation and conditional deallocation.

Two checks are justified rather than proved. Fetch does not write its
out parameter on the failure paths, since there is no value of the
private type T to write, and the API documents that callers check the
status first. The defensive assertion in Fetch that a stored value
deserializes successfully guards against corruption of the storage after
it was written, which no proof can rule out.

A prover package instantiates the generic so that GNATprove has an
instance to analyze, since generics are only analyzed at instantiation
points. It is referenced by nothing and adds no code to any build.
EOF
## Summary

Converts `src/data_structures/database/variable_database.ads/.adb` (the store beneath the product database component) to SPARK and proves it at the **silver** level with GNATprove at `--level=2`. Single commit, stacked on #217, which puts the serializer declarations in SPARK so this package can instantiate `Variable_Serializer` directly.

```
Total   103   0 unproved, 3 justified, 0 pragma Assume
```

## What changed

- **Ghost model.** `Is_Valid` (table allocated, bounds within `Id_Type`), `First_Id`/`Last_Id`/`Contains` as ghost accessors. Every operation preserves `Is_Valid`; the status contracts say an operation fails with `Id_Out_Of_Range` exactly when the Id is outside the held range, which the defensive assertion in `Clear_Override_All` relies on.
- **Proof-only contracts.** The usual `Assertion_Policy` block; generated code is unchanged. Functions with `out` parameters carry `Side_Effects`.
- **Serializer.** `T_Serializer` is instantiated directly in the private part as before and `Entry_Bytes` is its `Byte_Array`. `Store`/`Load` are ordinary SPARK code calling the instance. (Thanks to #217 no nested Off package or mirror subtype is needed.)
- **Excluded bodies.** `Init` and `Destroy` (heap allocation and conditional deallocation), with their postconditions documented.
- **Three justified checks.** `Fetch` leaves its `out` parameter unset on the two failure paths (no value of the private type `T` to write; callers check the status first), and the defensive assertion in `Fetch` that a stored value deserializes guards against storage corruption after writing, which no proof can rule out.
- `variable_database_prover.ads` instantiates the generic for analysis; referenced by nothing.

## Verification

- `redo prove`: 103 checks, 0 unproved. Package unit tests pass, style clean.
- Product database component test suite passes on the fork PR (dinkelk/adamant#183, which this supersedes).

### DIFF
--- a/src/data_structures/database/all.prove.yaml
+++ b/src/data_structures/database/all.prove.yaml
@@ -1,0 +1,4 @@
+---
+description: GNATprove configuration for the database packages. The target is silver, absence of runtime errors, including proof of the defensive assertions.
+level: 2
+mode: "silver"
--- a/src/data_structures/database/variable_database.adb
+++ b/src/data_structures/database/variable_database.adb
@@ -1,13 +1,49 @@
 with Safe_Deallocator;
 
-package body Variable_Database is
+package body Variable_Database with SPARK_Mode => On is
 
-   procedure Init (Self : in out Instance; Minimum_Id : in Id_Type; Maximum_Id : in Id_Type) is
+   -- See the note in the package specification. All contracts and ghost code
+   -- are for proof only and are disabled at runtime:
+   pragma Assertion_Policy
+      (Pre => Ignore,
+       Pre'Class => Ignore,
+       Post => Ignore,
+       Post'Class => Ignore,
+       Contract_Cases => Ignore,
+       Ghost => Ignore,
+       Loop_Invariant => Ignore,
+       Loop_Variant => Ignore,
+       Assert_And_Cut => Ignore,
+       Assume => Ignore,
+       Subprogram_Variant => Ignore);
+   pragma Unevaluated_Use_Of_Old (Allow);
+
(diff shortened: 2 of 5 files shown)


dinkelk:
Opened by mistake, this stays on the fork until approved.

PR #219, opened 2026-08-29
Single-source the CI dependency refs

dinkelk:
## Summary

Collects the CI dependency values duplicated across workflows into a single reusable workflow, `.github/workflows/_refs.yml`, consumed by every other workflow via `workflow_call` outputs. Today the only shared value is the toolchain container image tag (`ghcr.io/lasp/adamant:0.2`), which was repeated in all ten workflow files; a bump now edits exactly one line. Future cross-repo refs (for example a pinned sibling ref for one-off integration testing) land in the same file, where a temporary pin is an obvious one-line diff instead of hiding in a workflow env block.

The outputs are literal strings on the job rather than values written to `$GITHUB_OUTPUT` by a step: a `workflow_call` output must reference a job output, and routing an empty value through `$GITHUB_OUTPUT` has been observed to go stale across job re-runs.

### DIFF
--- a/.github/workflows/_refs.yml
+++ b/.github/workflows/_refs.yml
@@ -1,0 +1,36 @@
+# Single source of truth for the toolchain image (and any future
+# cross-repo refs) that this repo's CI builds against.
+#
+# Edit the "outputs:" block at the bottom. Every workflow in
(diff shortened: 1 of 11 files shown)


dinkelk:
Closing, decided against adopting this pattern here.

PR #221, opened 2026-09-01
Enable cross-compile of unit tests for bareboard targets under Renode

dinkelk:
Enables adamant's unit test suites to be cross-compiled for bareboard targets and run under Renode, and adds a workflow that runs the whole sweep as a PR check. This builds on the fixture state snapshot and Tester_Allocator work now merged to main.

Please review commit by commit to retain sanity.


gen/test/active_no_queue/test/test.adb
```diff
@@ -14,4 +14,6 @@ procedure Test is
    pragma Unreferenced (Tester);
 begin
    Put_Line ("If this compiled, it worked.");
+   --  Sentinel for the cross test runner (manual-test pass marker).
```
dinkelk: For all of these comments remove (manual-test pass marker)

src/data_structures/fifo/test/test.adb
```diff
@@ -85,4 +85,10 @@ begin
    Put ("Destroy queue test... ");
    My_Queue.Destroy;
    Put_Line ("passed.");
+
+   --  Sentinel for the cross test runner. The renode test rule
+   --  matches this case-insensitively as a manual-test pass marker
+   --  (since this test is not an AUnit suite and doesn't emit
```
dinkelk: simple comment only, Sentinel for the cross test runner.

gen/templates/tests/component-name_reciprocal.adb
```diff
@@ -474,7 +474,17 @@ package body Component.{{ name }}_Reciprocal is
       declare
          use Serializer_Types;
          Item : {{ p.type }};
+         --  Zero-fill the whole Item record before any partial copy.
+         --  Without this, when deserialization fails (e.g. a
+         --  packetizer truncated the input), only the first
+         --  P.Header.Buffer_Length bytes get overlaid and the trailing
+         --  bytes of Item retain whatever was on the stack -- showing
+         --  up as 0xFF / garbage in test assertions and in any
+         --  Representation.To_Tuple_String log output.
```
dinkelk: Simple comment only: "Zero-fill the whole Item record before any partial copy."

src/components/logger/test_variable/variable_tests-implementation.adb
```diff
@@ -25,14 +25,22 @@ with Interfaces; use Interfaces;
 
 package body Variable_Tests.Implementation is
 
+   -- Target-aware tester allocator: heap on Linux, a static instance on
+   -- bareboard, where Free also restores the Tester's fresh state.
+   -- Generic components must instantiate this themselves; the generated
+   -- unit test base cannot (it does not know the concrete types).
+   package Tester_Alloc is new Tester_Allocator
+     (Tester_Inst   => Component_Tester_Package.Instance,
+      Tester_Access => Component_Tester_Package.Instance_Access);
```
dinkelk: All the spacing on these is weird. We usually do 
```
Tester_Allocator (
   Tester_Inst
   ...
);
```
fix for all.

gen/test/generic_queued/test/test.adb
```diff
@@ -34,4 +29,6 @@ begin
 
    Tester.Final_Base;
    Comp.Final_Base;
+   --  Sentinel for the cross test runner (manual-test pass marker).
+   Put_Line ("=== ALL TESTS PASSED ===");
```
dinkelk, line 33: this needs to go in the sentinel commit, not here.

src/data_structures/variable_queue/test/global.ads
```diff
@@ -18,7 +18,10 @@ package Global is
    end Action;
    The_Action : Action;
 
-   task Unblock;
+   --  The default task stack is small on some bareboard runtimes, and
+   --  the blocking-error probes propagate an exception whose raise and
+   --  traceback machinery alone needs several kilobytes of stack.
+   task Unblock with Storage_Size => 16_384;
```
dinkelk, line 24: Please split this commit into 3.

1. For storage size and delay adjustments
2. Simulator refactors
3. Protected object hoisting.

As is, this commit is too large to review.

src/types/sys_time/test/sys_time_tests-implementation.adb
```diff
@@ -144,7 +144,9 @@ package body Sys_Time_Tests.Implementation is
       -- Put_Line (ASCII.LF & "Difference Sum(nsec): " & Float'Image( diff_sum_float ));
       Diff_Ave := (Diff_Sum_Float / Float (Tests)) * 10.0**9;
       Put_Line (ASCII.LF & "Add Time Random Tests" & ", Tests Run:" & Integer'Image (Tests));
-      Put_Line ("   Difference Max(nsec):   " & Integer'Image (Diff_Max / Nanoseconds (1)));
+      -- Print via Duration: dividing by Nanoseconds (1) divides by zero
```
dinkelk: this comment is immediately stale. just remove, no comment needed.

src/types/sys_time/test/sys_time_tests-implementation.adb
```diff
@@ -230,7 +235,9 @@ package body Sys_Time_Tests.Implementation is
       -- Put_Line (ASCII.LF & "Difference Sum(nsec): " & Float'Image( diff_sum_float ));
       Diff_Ave := (Diff_Sum_Float / Float (Tests)) * 10.0**9;
       Put_Line (ASCII.LF & "Subtract Time Random Tests" & ", Tests Run:" & Integer'Image (Tests));
-      Put_Line ("   Difference Max(nsec):   " & Integer'Image (Diff_Max / Nanoseconds (1)));
+      -- Print via Duration: dividing by Nanoseconds (1) divides by zero
+      -- on targets whose clock tick is coarser than a nanosecond.
```
dinkelk: stale comment, remove

src/types/sys_time/test/sys_time_tests-implementation.adb
```diff
@@ -318,7 +325,9 @@ package body Sys_Time_Tests.Implementation is
       -- Put_Line (ASCII.LF & "Difference Sum(nsec): " & Float'Image( diff_sum_float ));
       Diff_Ave := (Diff_Sum_Float / Float (Tests)) * 10.0**9;
       Put_Line (ASCII.LF & "Subtract Time Random Time Span Tests" & ", Tests Run:" & Integer'Image (Tests));
-      Put_Line ("   Difference Max(nsec):   " & Integer'Image (Diff_Max / Nanoseconds (1)));
+      -- Print via Duration: dividing by Nanoseconds (1) divides by zero
+      -- on targets whose clock tick is coarser than a nanosecond.
```
dinkelk: stale comment remove

src/components/sequence_store/test/sequence_store_tests-implementation.adb
```diff
@@ -1409,8 +1409,10 @@ package body Sequence_Store_Tests.Implementation is
       Cmd : Command.T;
       Seq_Region : constant Memory_Region.T := (Address => Slot_0_Memory'Address, Length => 0);
    begin
-      -- Send 3 commands to fill up queue.
-      Cmd.Header.Arg_Buffer_Length := Cmd.Arg_Buffer'Length;
+      -- Send 3 commands to fill up queue. Set an Id the tester does not
+      -- recognize. On bareboard, Initialize_Scalars zeroes the Id, and an
+      -- Id below Id_Base underflows in the tester's command logging.
```
dinkelk: None of the comments in this entire commit are necessary. Please remove them all.

src/components/fault_correction/test/fault_correction_tests-implementation.adb
```diff
@@ -88,17 +90,22 @@ package body Fault_Correction_Tests.Implementation is
       end Init_Duplicate;
 
       procedure Init_Too_Many is
+         --  Each config is ~270 bytes, so a 5001-element literal would
+         --  overflow the bareboard task stack. Allocate it on the heap
+         --  instead and free it when the scenario is done.
```
dinkelk: Simplify comment to just say "Allocate on heap to avoid overflowing task stack, especially on bareboard targets."

src/components/cpu_monitor/test/bb/cpu_monitor_test_interrupts.adb
```diff
@@ -0,0 +1,7 @@
+package body Cpu_Monitor_Test_Interrupts is
+
+   List : aliased Interrupt_Types.Interrupt_Id_List := [];
```
dinkelk, line 6: Need comment saying that this is empty because there is no hook to uniformly stimulate interrupts on all bareboard targets, so we only test this feature on Linux for unit level testing.

gen/test/records/test/.skip_test_renode
dinkelk, line 1: if this, or any other file in this is skipped to compiler bug, they need to be committed under THEIR OWN commit, so they are easily found and verified fix when a compiler fix arrives. Please audit and split those out.

doc/example_architecture/example_component/test/.skip_test_renode
dinkelk, line 1: Make no reference to specific closed source compiler versions like: GNAT Pro 25.2 riscv32-elf in commit messages.

gen/test/active_no_queue/test/test.adb
dinkelk: Done. Every sentinel comment now reads just "Sentinel for the cross test runner." across all 25 files.

src/data_structures/fifo/test/test.adb
dinkelk: Done. Trimmed to the same one-line comment the others use.

src/components/logger/test_variable/variable_tests-implementation.adb
dinkelk: Reformatted to the house style, matching what the repo already does elsewhere (for example the Safe_Deallocator instantiation in parameter_table_buffer.adb): open paren at end of line, arguments indented one level with no alignment padding, closing ");" on its own line.

Two notes. Two of the eight instances, limiter/test and limiter/test_variable, were already on main from the state snapshot PR with the same spacing, and I fixed those too since you said fix for all. Tell me if you would rather this PR touch only the six it introduces. I also wrapped the long Safe_Deallocator instantiation in the fault_correction test the same way, since it was a single 150 character line.

gen/test/generic_queued/test/test.adb
dinkelk: Moved into the sentinel commit. This commit now only does the Global hoist for that file.

src/data_structures/variable_queue/test/global.ads
dinkelk: Split into three commits, in the order you listed:

1. Give the blocking queue test tasks stack size and idle delays
2. Move the test simulator tasks into their own packages
3. Hoist local protected objects in tests to library level

src/types/sys_time/test/sys_time_tests-implementation.adb
dinkelk: Removed.

src/types/sys_time/test/sys_time_tests-implementation.adb
dinkelk: Removed.

src/types/sys_time/test/sys_time_tests-implementation.adb
dinkelk: Removed.

src/components/sequence_store/test/sequence_store_tests-implementation.adb
dinkelk: Done. All four comments that commit added are gone, and the two pre-existing "Send 3 commands to fill up queue." lines are back to their original wording.

src/components/fault_correction/test/fault_correction_tests-implementation.adb
dinkelk: Done, the comment now reads exactly that.

src/components/cpu_monitor/test/bb/cpu_monitor_test_interrupts.adb
dinkelk: Added to the bareboard body: empty because there is no hook to uniformly stimulate interrupts on all bareboard targets, so that feature is only tested on Linux for unit level testing.

gen/test/records/test/.skip_test_renode
dinkelk: Audited all 19 markers. gen/test/records is the only one caused by a compiler bug, and it now has its own commit, "Skip the records generator test that crashes the cross compiler", placed immediately before the dodge commit so both compiler items sit together at the end of the branch. Removing that one commit is the whole revert when a fixed compiler arrives.

The other 18 markers break down as: POSIX, socket and disk dependencies (9), one Jorvik violation, one test whose only verdict is a byte diff the runner cannot read, one Jinja placeholder that is not a real test, and the six tutorial tests that fail by design.

doc/example_architecture/example_component/test/.skip_test_renode
dinkelk: Removed. That commit message no longer names the compiler product, its version, or its internal function. It now says the cross compiler hits an internal error on a function that returns by value a record containing an array whose index range starts negative and crosses zero, reported to the vendor with a minimal reproducer.

One question: the dodge commit still says "riscv32" in its subject and body, as the architecture rather than a compiler version. I kept it because it is useful when someone revisits that workaround later, but I am happy to genericize it if you want the architecture out of commit messages too.

gen/templates/tests/component-name_reciprocal.adb
dinkelk: Done, the comment is now exactly that one line.

gen/templates/tests/component-name_reciprocal.adb
```diff
@@ -474,7 +474,11 @@ package body Component.{{ name }}_Reciprocal is
       declare
          use Serializer_Types;
          Item : {{ p.type }};
+         --  Zero-fill the whole Item record before any partial copy.
+         Item_As_Bytes : Basic_Types.Byte_Array (0 .. {{ p.type_package }}.Size_In_Bytes - 1)
+            with Import, Convention => Ada, Address => Item'Address;
       begin
+         Item_As_Bytes := [others => 0];
```
dinkelk: Need to go look at this. What tests fail for this, and are they just bad test assertions? Or is this actually a real fix?

src/components/parameter_store/test/parameter_store_tests-implementation.adb
```diff
@@ -541,7 +541,7 @@ package body Parameter_Store_Tests.Implementation is
       T.Component_Instance.Set_Up;
 
       -- Send 3 commands to fill up queue.
-      Cmd.Header.Arg_Buffer_Length := Cmd.Arg_Buffer'Length;
+      Cmd.Header := (Source_Id => 0, Id => 16#FFFF#, Arg_Buffer_Length => Cmd.Arg_Buffer'Length);
```
dinkelk: Does this instead indicate a bad assertion? Because this seems like a workaround for a bad check.

cbrisebois-lasp (approved):

Jbsco (approved):

gen/templates/tests/component-name_reciprocal.adb
dinkelk: You were right to push on this. I investigated and my commit message was overstating it, so I have dropped the commit from this PR.

What I found, with three experiments:

1. Reverted the zero-fill and ran all seven components that have this code path on the cross target. All seven pass. No test fails without it.
2. Instrumented the fallback branch to see whether it is even reached. It is: 3 hits in ccsds_command_depacketizer, 1 in ccsds_router, 5 in ccsds_subpacket_extractor.
3. Replaced the uninitialized bytes with a deterministic 16#FF# fill to stand in for stack garbage, so the "garbage" case is reproducible on the host. All three suites still pass, which proves no assertion reads those trailing bytes.

So it is not a bad assertion, and it is not fixing a failing test either. It removes real undefined behavior, since the fallback writes only Buffer_Length bytes and then the whole Item is both printed through To_Tuple_String and passed to the handler that records history, but the only observable effect today is log text. My commit claimed the garbage was "visible on cross targets in test assertions and Representation output", and the test assertions half of that is unsupported.

It does not belong in a PR about cross compiling the tests, so it is out. I kept the change on a local branch in case you want it as its own hardening change, with an honest message this time.

src/components/parameter_store/test/parameter_store_tests-implementation.adb
dinkelk: You were right, this was a workaround for a bad check. The real defect is in the generated tester.

Log_Incoming_Parameter guards both bounds before subtracting the id base:

    if Natural (Param.Header.Id) <= Num_Parameters and then Param.Header.Id >= Get_Id_Base (Self.Parameters) then

Log_Incoming_Command only guards the upper one:

    if Natural (Cmd.Header.Id) <= Num_Commands then

Command_Id is "new Natural range 0 .. 65_535", so an id below the id base underflows the subtraction. Nothing noticed on Linux because an uninitialized Command.T header holds a large garbage id that fails the upper bound and is skipped. On a target where Initialize_Scalars zeroes the header, the id is 0, passes the upper bound, and underflows.

So this commit is replaced by a one line template fix, "Guard the tester command log against ids below the id base", and all three test files (parameter_store, parameter_table_forwarder, sequence_store) are back to their original form untouched. Verified: those three pass on host and cross with the workaround gone, and I also ran four components whose generated tester the guard touches, host and cross, plus template style.

gen/templates/tests/component-name_reciprocal.adb
```diff
@@ -636,7 +636,7 @@ package body Component.{{ name }}_Reciprocal is
       Local_Id : {{ commands.name }}.Local_Command_Id_Type;
    begin
       if Self.Log_Commands then
-         if Natural (Cmd.Header.Id) <= {{ commands.name }}.Num_Commands then
+         if Natural (Cmd.Header.Id) <= {{ commands.name }}.Num_Commands and then Cmd.Header.Id >= {{ commands.name }}.Get_Id_Base (Self.Commands) then
```
dinkelk, line 639: Is there similar guards needed like this for other ided entities in this template?

dinkelk: I audited every id'd entity in this template. There are exactly two others that convert an id to a Local_..._Id_Type: Dispatch_Event at line 370 and Dispatch_Data_Product at line 411. Neither needs this guard, and the reason is in the defaults:

    commands     Id_Base default 1   subtracts the base, both bounds guarded
    parameters   Id_Base default 1   subtracts the base, both bounds guarded
    events       Id_Base default 0   no subtraction, so nothing to underflow
    data products Id_Base default 0  no subtraction, so nothing to underflow

The command and parameter paths subtract the base because their base starts at 1, and subtraction is what can go negative. Events and data products index the dispatch table with the raw id, so there is no arithmetic to guard. Parameters already carried both bounds, which is where I copied the fix from.

They are also different in kind. Log_Incoming_Command is fed a header the test itself builds, which is how it can see an uninitialized or zeroed id. The event and data product dispatchers are fed ids the component under test generated from its own id base, so a garbage id cannot reach them the same way.

One pre-existing assumption worth naming while we are here, though I have left it alone as out of scope: 'Val (Id) with no subtraction is only correct while those two bases stay 0. If a test ever set a nonzero Event_Id_Base or Data_Product_Id_Base, the dispatcher would index the wrong table entry or raise. Nothing exercises that today. The only bases any test sets are Parameter_Id_Base, in the parameters tests, at 1, 3 and 5, and that path is guarded.

PR #225, opened 2026-09-09
Add Register_Command helper to the component base package

owhite7128:
This should remove the need for the command sequencer to reach down into the component internals to registers its extra commands. This provides an abstraction layer that is built into the component package in case other components need this in the future.


dinkelk (commented):
Looks good!

dinkelk (approved):

PR #178, opened 2026-05-08
Add Simple Command Sequencer

owhite7128:
This component adds functionality to define command sequences in YAML that can be run through this component. It has support for simple time sleeping in sequences, a few useful flags to allow control of flow, and dynamic arguments that can be passed in when the sequence is run instead of at compile time.

PR #227, opened 2026-09-10
Add Simple Command Sequencer (merge)

owhite7128:
This is the same code as #178 . Discussion lives there, this is just for eventual merging. I will flatten the history here as to preserve a clean git history


src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -0,0 +1,593 @@
...
+                  -- else is late or stale (timed out, killed, or the frame was
+                  -- reused) and is ignored: acting on it would advance the wrong step.
+                  if Frame.Status = Waiting_For_Cmd_Resp and then Arg.Command_Id = Frame.Pending_Command_Id then
+                     if Arg.Status = Command_Response_Status.Failure then
+                        Self.Event_T_Send_If_Connected (Self.Events.Command_Failure (Time,
+                           (Sequence_Id => Frame.Sequence_Id, Frame_Id => Frame_To_Wake_Id,
+                            Step => Frame.Step, Command_Id => Arg.Command_Id)));
+                     end if;
+
+                     if Arg.Status = Command_Response_Status.Failure and then Seq.Abort_On_Failed_Cmd then
```
dinkelk, line 338: Better to check for not success. Commands can fail for other reasons than just "Failure". Here is a different idea:

```
   declare
      Failed : constant Boolean := Arg.Status /= Command_Response_Status.Success;
   begin
      if Failed then
         -- Any non-success status is a failed sub-command: Failure, Id_Error,
         -- Validation_Error, Length_Error, or Dropped.
         Self.Event_T_Send_If_Connected (Self.Events.Command_Failure (Time, ...));
      end if;

      if Failed and then Seq.Abort_On_Failed_Cmd then
         Finish_Sequence (Self, Frame, Command_Response_Status.Failure, Time);
         Self.Event_T_Send_If_Connected (Self.Events.Sequence_Aborted (Time, ...));
      else
         Frame.Status := Running;
         Execute_Sequence (Self, Frame);
      end if;
   end;
```

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -0,0 +1,202 @@
...
+                      using "Arg.field_name" syntax. If the command takes no
+                      argument, omit this field. Only valid alongside `command:`.
+
+                      Examples:
+                        - "Arg" - pass through the entire sequence argument
+                        - "Arg.Power_Mode" - pass a field from the sequence argument
+                        - "(Mode => Arg.Mode, Enable => True)" - construct a record
+                        - "42" - a literal value
+
+                  wait_for_completion:
```
dinkelk, line 167: This appears unused. Either we need to remove or implement properly. What do you think? Seems like this is a nice feature to have to me. I am not sure if we decided to scrap it and it didn't get fully scrapped. There are no unit tests for it either.

Idea from AI:

Issue: The schema documents a per-step override, the model computes step.wait_for_completion, and then nothing consumes it. Step has no wait field, Sequence_Type.Wait_For_Cmd_Resp is per sequence, and Dispatch_Step_Command (component-simple_command_sequencer-implementation.adb:141) reads only the sequence-level flag. A sequence author who marks one step wait_for_completion: true inside a no-wait sequence gets no response wait and no timeout protection for that step. The reverse case (a false step inside a waiting sequence) blocks on a response the author did not expect to wait for. Either behavior diverges silently from the model the author wrote. No unit test exercises the override, so the gap is invisible.

Proposed Fix: Implement the override, since the schema already promises it. Add the flag to the step record and use it at dispatch:

```
-- types/simple_sequencer_types.ads
type Step (Kind : Step_Kind := Command_Step) is record
   Id            : Command_Types.Command_Id := 0;
   Arg_Length    : Command_Types.Command_Arg_Buffer_Length_Type := 0;
   -- Park the frame for this step's response (and arm its timeout).
   Wait_For_Resp : Boolean := True;
   case Kind is
      ...
```

```
{# gen/templates/command_sequences/name.ads, every command-step form #}
(Kind          => Command_Step,
 Id            => {{ step.component_name }}_{{ step.command_name }},
 Wait_For_Resp => {{ "True" if step.wait_for_completion else "False" }},
 ...
```

```
-- component-simple_command_sequencer-implementation.adb, Dispatch_Step_Command
procedure Dispatch_Step_Command (Self : in out Instance; Frame : in out Sequence_Frame; Seq : in Sequence_Type; Step_Obj : in Step; Cmd : in Command.T; Time : in Sys_Time.T) is
begin
   if Step_Obj.Wait_For_Resp then
      Add_Status := Add (Time, Seq.Command_Timeout, Frame.Timeout_Deadline);
      ...
```

Sequence_Type.Wait_For_Cmd_Resp then becomes redundant and can be dropped. If the override is instead not wanted, remove wait_for_completion from the schema and have sequence_step.validate raise a ModelException when it is present, so the model cannot describe behavior the sequencer does not have.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.ads
```diff
@@ -0,0 +1,131 @@
...
+   -- Tick for managing timeouts and delays
+   overriding procedure Tick_T_Recv_Async (Self : in out Instance; Arg : in Tick.T);
+   -- This procedure is called when a Tick_T_Recv_Async message is dropped due to a full queue.
+   overriding procedure Tick_T_Recv_Async_Dropped (Self : in out Instance; Arg : in Tick.T);
+
+   ---------------------------------------
+   -- Invoker connector primitives:
+   ---------------------------------------
+   -- This procedure is called when a Command_T_Send message is dropped due to a full queue.
+   overriding procedure Command_T_Send_Dropped (Self : in out Instance; Arg : in Command.T) is null;
```
dinkelk, line 94: I think we should override this. A command that fails to dispatch at the router is reported via timeout. But we can short circuit that and report more accurately. Consider something like:

```
overriding procedure Command_T_Send_Dropped (Self : in out Instance; Arg : in Command.T) is
   Time : constant Sys_Time.T := Self.Sys_Time_T_Get;
begin
   Self.Event_T_Send_If_Connected (Self.Events.Dropped_Outgoing_Command (Time, Arg.Header));
   -- The router will never answer a command it never received; resolve the
   -- waiting frame now as if the router had replied Dropped.
   Self.Command_Response_T_Recv_Async ((Source_Id => Arg.Header.Source_Id, Registration_Id => Self.Command_Reg_Id, Command_Id => Arg.Header.Id, Status => Command_Response_Status.Dropped));
end Command_T_Send_Dropped;
```

src/components/simple_command_sequencer/gen/doc/simple_command_sequencer_generator.tex
```diff
@@ -0,0 +1,78 @@
...
...g.\ \texttt{Sequence\_B}) takes the sequence's own \texttt{arg\_type} verbatim as its command argument (or no argument for an argless sequence) -- a user-written, normally-registered type, so no argument records are generated or registered anywhere in the pipeline. At run time the component copies the received argument buffer unchanged into a generic \texttt{Run\_Sequence\_Arg.T} and dispatches through the existing \texttt{Run\_Sequence} backbone using the sequence's runtime slot index. \texttt{Run\_Sequence} rejects any argument whose length differs from the serialized length of the sequence's \texttt{arg\_type} (recorded in the autocoded table) with an \texttt{Invalid\_Sequence\_Argument\_Length} event; the response behavior comes from the sequence's static configuration in the same table (see Sequence Response Behavior below). A companion generated package, \texttt{<suite>\_Commands}, carries the operator-side builder surface (per-sequence command id getters and \texttt{Command.T} constructors) for unit tests and other on-board callers -- split from the sequences package because the sequencer component itself never uses the builders. No per-sequence Ada needs to be hand-written:
```
dinkelk, line 49: I don't think you use this discovery model anymore. Audit this. You use init param now.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -0,0 +1,202 @@
...
+      sizes the suite's generated summary packet ground type (one
+      Sequence_Frame_Summary per frame), so the frame pool and the packet
+      layout can never disagree. The upper bound -- how many frame
+      summaries fit in one packet buffer -- is enforced by the model
+      against the project configuration.
+
+  command_timeout_seconds:
+    required: false
+    type: int
+    range: { min: 0, max: 2147483 }
```
dinkelk, line 42: Should we really allow zero here? Maybe we should restrict to min of 1. Or should we make zero mean we wait forever instead? I think we should do one of the two.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -0,0 +1,202 @@
...
+    desc: |
+      Default timeout, in whole seconds, that the sequencer will wait for
+      a sub-command's response before declaring the sequence failed. Used
+      for any sequence in this suite that does not specify its own
+      command_timeout_seconds. Bounded so the millisecond count always
+      fits an Ada.Real_Time.Time_Span.
```
dinkelk, line 48: Is this is seconds or milliseconds? You mention milliseconds here, so now I am concerned. Milliseconds seems better. 1 second could be seen as too long of a timeout for some systems using this (unlikely but possible)

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -0,0 +1,202 @@
...
+            enum: ['send_after_sequence_start', 'send_after_sequence_completion']
+            desc: |
+              When the sequencer replies to this sequence's own command:
+              send_after_sequence_start (the default) acknowledges as soon as
+              the sequence is started; send_after_sequence_completion defers
+              the reply until the sequence finishes, carrying its final
+              success or failure. Static per-sequence configuration; the
+              generic Run_Sequence command uses it too.
+
+          command_timeout_seconds:
```
dinkelk, line 121: same note here as for the global command_timeout_seconds (see above)

src/components/simple_command_sequencer/types/simple_sequencer_types.ads
```diff
@@ -0,0 +1,109 @@
...
+      Num_Concurrent_Sequences : Num_Concurrent_Sequences_Type;
+   end record;
+
+   -- One running sequence. Internal state only, never serialized; the
+   -- downlinked view is Sequence_Frame_Summary. Run_Sequence re-seeds every
+   -- per-run field when it claims a frame, and ending a sequence only sets
+   -- Status to Not_Running, so an idle frame still reports its last run.
+   type Sequence_Frame is record
+      Sequence_Id : Interfaces.Unsigned_16 := 0;
+      Frame_Id : Interfaces.Unsigned_32 := 0;
```
dinkelk, line 85: Unsigned_32 is too large here. We already have a bound we can use computed better typed range above:

```
subtype Frame_Id_Type is Interfaces.Unsigned_16 range 0 .. Num_Concurrent_Sequences_Type'Last - 1
```

Additional benefit is we can make the following smaller:

Packed_U16.T for Invalid_Frame_Id and Frame_Not_Running. The event records' Frame_Id fields can then be U16 as well.

src/components/simple_command_sequencer/types/simple_sequencer_types.ads
```diff
@@ -0,0 +1,109 @@
...
+      -- Copied from the sequence table at claim time for the summary packet.
+      Response_Behavior : Sequence_Enums.Sequence_Response_Behavior.E := Sequence_Enums.Sequence_Response_Behavior.Send_After_Sequence_Start;
+      -- Operator response context, used by the deferred-reply paths:
+      Operator_Source_Id : Command_Types.Command_Source_Id := 0;
+      Operator_Command_Id : Command_Types.Command_Id := 0;
+      -- The caller's argument buffer, traversed by the Resolvers of dynamic steps.
+      Dynamic_Arg : Run_Sequence_Buffer_Type := [others => 0];
+   end record;
+
+   type Sequence_Frame_Array is array (Interfaces.Unsigned_32 range <>) of Sequence_Frame;
```
dinkelk, line 107: Index by `Frame_Id_Type` per my suggestion above.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -0,0 +1,593 @@
...
+      if Arg.Status = Command_Response_Status.Register_Source then
+         declare
+            Source_Id_Set : Boolean := False;
+         begin
+            for Frame of Self.Sequence_Frames.all loop
+               if Frame.Has_Source_Id = False then
+                  Frame.Source_Id := Arg.Source_Id;
+                  Frame.Has_Source_Id := True;
+                  Source_Id_Set := True;
+                  exit;
```
dinkelk, line 311: One thing about this logic. It does not preclude the fact that two frames can have the same source ID. This could result from a miswiring, and could have bad side-effects and misrouting of command responses.

I think we need to check that  the Arg.Source_Id is not held by another frame before assigning it. for example:

```
declare
   Existing_Frame : Interfaces.Unsigned_32;
begin
   if Find_Sequence_Frame_Id_From_Source_Id (Self, Arg.Source_Id, Existing_Frame) then
      -- Already registered: a duplicate id would shadow the existing frame.
      Self.Event_T_Send_If_Connected (Self.Events.Duplicate_Register_Source (Time, (Value => Source_Id));
   else
      for Frame of Self.Sequence_Frames.all loop
         if not Frame.Has_Source_Id then
            ...
```

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -0,0 +1,593 @@
...
+         end if;
+         Frame.Pending_Command_Id := Cmd.Header.Id;
+         Frame.Status := Waiting_For_Cmd_Resp;
+      end if;
+      Self.Command_T_Send (Cmd);
+      Note_Command_Sent (Self, Time);
+   end Dispatch_Step_Command;
+
+   -- Run `Frame` until it parks (command response or sleep) or its sequence ends.
+   procedure Execute_Sequence (Self : in out Instance; Frame : in out Sequence_Frame) is
```
dinkelk, line 156: This does a lot of fetching of sys time. But we could probably just pass sys_time in here from caller. Same pattern we use elsewhere. Better efficiency and less sys_time calls.

dinkelk (changes requested):

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.ads
dinkelk: nvm, i think this is OK based on command router design, which basically does exactly this, so we don't need to duplicate that here.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: refactored, updated tests

src/components/simple_command_sequencer/gen/doc/simple_command_sequencer_generator.tex
cbrisebois-lasp: audited

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: restricting to 1

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: restricted to 1

src/components/simple_command_sequencer/types/simple_sequencer_types.ads
cbrisebois-lasp: adopted

src/components/simple_command_sequencer/types/simple_sequencer_types.ads
cbrisebois-lasp: indexed

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: parametrized

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: implementing check at registration. All frames will have a unique source id

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: implemented

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
dinkelk: why is this one not changed too?

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -4,11 +4,13 @@
 type: map
 mapping:
   name:
-    required: true
+    required: false
```
dinkelk: yeah, this is weird, nonstandard for all of adamant. package names are ALWAYS derived from the filename... so i think we should simply remove this.

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -671,24 +675,39 @@ def _check_engine_connection_counts(self):
...
+                    f"component '{comp.instance_name}' receives "
+                    "Command_Response_T_Recv_Async connections from multiple "
+                    f"components ({', '.join(sorted(senders))}). Each sender "
+                    "numbers its Register_Source ids from its own connector "
+                    "indices, so ids from different senders can collide; the "
+                    "component refuses a duplicate at runtime "
+                    "(Duplicate_Register_Source), leaving that frame "
+                    "unclaimable. Route every connection through a single "
+                    "command router."
+                )
```
dinkelk: this check should go away.

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -378,6 +378,25 @@ def __init__(
...
+                if later.command == step.command and later.wait_for_completion:
+                    raise ModelException(
+                        f"Sequence '{self.name}': step {idx} dispatches "
+                        f"'{step.command}' without waiting for completion, "
+                        f"and step {later.index} waits on the same command. "
+                        "The no-wait step's late response could wake the "
+                        "waiting step early, since responses are matched by "
+                        "command id. Reorder the steps, use a different "
+                        "command, or make both steps wait."
+                    )
```
dinkelk: This whole issue makes me want to just simplify and remove the feature altogether. I feel like maybe this came up before and the feature was removed from everywhere except the schema and parsing.

I don't like the complexity added here. This is the simple sequencer afterall. It seems like users probably can get away with wait for completion behavior on entire sequence or not. And that makes writing the sequence more straightfoward (e.g. do i need to use sleeps to space things out, or not, etc).

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -0,0 +1,212 @@
...
+          wait_for_command_completion:
+            required: false
+            type: bool
+            default: true
+            desc: |
+              Default setting for whether to wait for each sub-command to
+              complete before executing the next one. Individual steps can
+              override this.
+
+          continue_on_failure:
```
dinkelk, line 105: One interesting thing. I think we need to not allow this to be False if wait_for_command_completion is False. If you are not waiting for responses, you can't take ANY action on failure.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -0,0 +1,212 @@
...
+                      using "Arg.field_name" syntax. If the command takes no
+                      argument, omit this field. Only valid alongside `command:`.
+
+                      Examples:
+                        - "Arg" - pass through the entire sequence argument
+                        - "Arg.Power_Mode" - pass a field from the sequence argument
+                        - "(Mode => Arg.Mode, Enable => True)" - construct a record
+                        - "42" - a literal value
+
+                  wait_for_completion:
```
dinkelk: I am now thinking we just remove this feature. And only add it back when we have a use case that warrants it.

dinkelk (changes requested):

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: removed

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: fixing

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: let's talk about this, see below

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: let's talk about this, I think it's already done. And I don't see why we should remove it if it works now

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: good catch, adding a model check

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: good point removed

src/components/simple_command_sequencer/gen/models/command_sequences.py
dinkelk: Because of this, I vote we drop this feature in favor of a simpler design where user must chose wait for response or fire and forget at the sequence level. Simple, easier to reason about, less bugs I hope.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -102,9 +103,33 @@ package body Component.Simple_Command_Sequencer.Implementation is
...
-   -- Return `Frame` to idle, update the counters and data products, and emit
-   -- any deferred reply. Only Status is reset: the next claim re-seeds the
-   -- rest, and while idle it lets the summary packet report the last run.
+   -- Park `Frame` in Draining and stamp its deadline one sequence
+   -- Response_Timeout from now. The deadline bounds inactivity, not the whole
+   -- drain -- it is restarted whenever an outstanding response arrives -- so a
+   -- permanently lost response cannot leak the frame. On Sys_Time overflow the
+   -- frame is released immediately: the stale-response window is preferable to
+   -- a frame parked on a deadline that never passes.
+   procedure Park_Draining (Self : in out Instance; Frame : in out Sequence_Frame; Time : in Sys_Time.T) is
```
dinkelk: The AI made up terminology slop is killing me here man. WTF does a park frame drain mean. Can we make this plain english

src/components/simple_command_sequencer/types/simple_sequencer_types.ads
```diff
@@ -55,6 +55,11 @@ package Simple_Sequencer_Types is
       Abort_On_Failed_Cmd   : Boolean;
       -- Converted from the model's milliseconds once, at elaboration.
       Command_Timeout       : Ada.Real_Time.Time_Span;
+      -- Outstanding-response timeout: how long a frame whose sequence has
+      -- ended may await its outstanding sub-command responses without any
+      -- arriving before the remainder are given up for lost and the frame is
+      -- released.
+      Response_Timeout      : Ada.Real_Time.Time_Span;
```
dinkelk: This is literally the same definition as Command_Timeout, so I suggest we just use that.

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -305,6 +305,7 @@ def __init__(
         wait_for_command_completion=True,
         continue_on_failure=False,
         command_timeout_seconds=None,
+        response_timeout_seconds=None,
```
dinkelk: Having a hard time with this commit. The complexity add is significant for a rare corner case.  I really think this whole issue can be 99% fixed with a much simpler solution:
1. Round robin
2. Just wait command_response timeout at the end of a sequence that is not in command_completion mode.

And 2 may not be necessary either. All of this is a bit artificial. It might be better to leave it up to the user. Like bro, put a sleep at the end of you sequence if you are firing off a final command that has a response that comes back really fast.

src/components/simple_command_sequencer/test/component-simple_command_sequencer-implementation-tester.adb
```diff
@@ -38,6 +38,7 @@ package body Component.Simple_Command_Sequencer.Implementation.Tester is
       Self.Dropped_Command_History.Init (Depth => 100);
       Self.Dropped_Command_Response_History.Init (Depth => 100);
```
dinkelk, line 39: This commit is not needed. The command router does this for us.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -90,7 +90,10 @@ mapping:
             type: str
             desc: |
               Name of this sequence. This will become a command that can be
-              executed. Must follow Ada naming conventions.
+              executed on the sequencer instance, so it may not reuse the name
+              of one of the sequencer's built-in commands (Run_Sequence,
+              Kill_All_Sequences, Kill_Frame, Set_Summary_Packet_Period); the
```
dinkelk: obviously. this is not needed here. This is just AI injecting noise.

src/components/simple_command_sequencer/gen/models/simple_command_sequencer_commands.py
```diff
@@ -74,6 +74,20 @@ def set_assembly(
         # Provide the assembly to the product packetizer model
         self.command_sequences_model.set_assembly(assembly)
 
+        # Each sequence becomes a command on the sequencer instance, so a sequence
```
dinkelk, line 77: this commit can stay if its cleaned up a bit.

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -548,6 +577,26 @@ def load(self):
                     lineno=seq.lineno,
                 )
 
+        # A sequence's argument travels inside the Run_Sequence command argument,
+        # behind its Sequence_Id and Arg_Length header fields, so it has less room
+        # than an ordinary command argument. The generated Sequences_Table stores
+        # each argument's serialized length in the passthrough length subtype,
+        # which would otherwise fail its range check at elaboration.
+        max_arg_bytes = get_max_sequence_arg_size_bytes()
```
dinkelk: I am not sure we need all this complexity. the ada type system checks this perfectly fine with way less code to maintain.

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -315,7 +315,7 @@ def __init__(
         description=None,
         arg_type=None,
         wait_for_command_completion=True,
-        continue_on_failure=False,
+        continue_on_failure=None,
```
dinkelk, line 296: this commit is a good change.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -3,15 +3,6 @@
...
 mapping:
-  name:
-    required: false
-    type: str
-    desc: |
-      Optional human-readable name for this command sequences suite. The
-      generated Ada package name is NOT taken from here -- it is derived
-      from the model filename
-      (<model_name>_Command_Sequences[_<specific_name>]).
-
```
dinkelk: this commit is a good change.

dinkelk (changes requested):

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: refactored to milliseconds in following commit

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -14,7 +14,7 @@
 from util import model_loader
 import re
 
-DEFAULT_COMMAND_TIMEOUT_SECONDS = 30
```
dinkelk: I think this is a win. One thing to audit, there are no changes to the implementation here? that is surprising to me.

src/components/simple_command_sequencer/test/simple_command_sequencer_tests-implementation.adb
```diff
@@ -389,13 +389,11 @@ package body Simple_Command_Sequencer_Tests.Implementation is
       Natural_Assert.Eq (T.Command_Failure_History.Get_Count, 1);
       Natural_Assert.Eq (T.Sequence_Aborted_History.Get_Count, 1);
 
-      -- The failure response already aborted the sequence and set the frame to Not_Running
```
dinkelk: I saw claude point this out to me too, and I didn't like the resulting code because it was non-obvious when steps were incremented. Could be swayed, or maybe there is a better way to write the code.

src/components/simple_command_sequencer/test/component-simple_command_sequencer-implementation-tester.adb
dinkelk, line 36: I vote we drop this commit.

src/components/simple_command_sequencer/gen/models/simple_command_sequencer_commands.py
dinkelk, line 77: easy win.

gen/models/submodels/ided_suite.py
```diff
@@ -259,6 +259,15 @@ def _initialize(
                 + "provided for some entities but not others, which is forbidden."
             )
 
+        self.derive_entity_lists()
```
dinkelk, line 262: this is probably a good change, but i need to review more thoroughly

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -408,11 +408,12 @@ def __init__(
         # frame has moved on and responses are matched by command id alone. A
         # later waiting step on the same command could therefore be woken
         # early by the no-wait step's stale response, so reject that shape.
+        # Ada names are case-insensitive, so the commands are compared that way.
         for idx, step in enumerate(self.steps):
             if step.command is None or step.wait_for_completion:
                 continue
             for later in self.steps[idx + 1:]:
-                if later.command == step.command and later.wait_for_completion:
+                if later.command is not None and later.command.lower() == step.command.lower() and later.wait_for_completion:
```
dinkelk: good change

src/components/simple_command_sequencer/gen/templates/command_sequences/name_commands.adb
```diff
@@ -10,6 +10,10 @@ package body {{ name }}_Commands is
 
    procedure Set_Id_Base (Self : in out Instance; Id_Base : in Command_Types.Command_Id) is
    begin
+      -- ID base set too high for the sequencer's command set: the static
+      -- commands followed by this suite's sequences. This is checked in the
+      -- assembly model.
+      pragma Assert (Natural (Id_Base) + Simple_Command_Sequencer_Commands.Num_Commands + {{ sequences|length }} - 1 <= Natural (Command_Types.Command_Id'Last));
```
dinkelk, line 16: this change is fine.

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -325,7 +325,10 @@ def __init__(
         self.description = description
         self.arg_type = arg_type
         self.wait_for_command_completion = wait_for_command_completion
-        self.continue_on_failure = continue_on_failure
+        # None means the key was omitted: the sequence aborts on failure, but
+        # the author promised nothing.
+        self.continue_on_failure = bool(continue_on_failure)
+        self._continue_on_failure_explicit = continue_on_failure is not None
```
dinkelk, line 308: this is a good change

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -527,16 +527,12 @@ def load(self):
...
         self.sequences = OrderedDict()
 
-        self.suite_name = None
         self.assembly_name = None
 
         self.name = ada.formatType(self.model_name) + "_Command_Sequences"
         if self.specific_name:
             self.name = self.name + "_" + ada.formatVariable(self.specific_name)
 
-        if "name" in self.data:
```
dinkelk: good change

dinkelk (changes requested):

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: I hear ya on simplification. but I the thing is, I instantly have a use case for this feature in DAS configuration. Some of the command have a direct response while others don't. I either have to have no confirmation or use false confirmation for the ones that don't have it.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: lol this is actually something I wanted haha

owhite7128:
AI added that frame "park draining" stuff previously, which while I agree the name is slop. I think it is important. I don't think right now that if we kill a sequence and a command response is still pending. there is nothing to reject that erroneous command response from impacting a sequence loaded later in the same frame. This feels like a pathway we want to prevent, but I don't see that that made it in. We should chat about that in the merge party

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: feature removed

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: feature removed. I actually told it to remove drain as well this must have slipped through

src/components/simple_command_sequencer/types/simple_sequencer_types.ads
cbrisebois-lasp: feature removed

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: round robin implement, other feature removed

src/components/simple_command_sequencer/test/component-simple_command_sequencer-implementation-tester.adb
cbrisebois-lasp: removed

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: removed

src/components/simple_command_sequencer/gen/models/simple_command_sequencer_commands.py
cbrisebois-lasp: cleaned up

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: removed

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: done

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: done

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: yeah pretty neat

src/components/simple_command_sequencer/test/simple_command_sequencer_tests-implementation.adb
cbrisebois-lasp: removed

src/components/simple_command_sequencer/test/component-simple_command_sequencer-implementation-tester.adb
cbrisebois-lasp: dropped

src/components/simple_command_sequencer/gen/models/simple_command_sequencer_commands.py
cbrisebois-lasp: won

gen/models/submodels/ided_suite.py
cbrisebois-lasp: yeah I kept this in. This is the one discussion item for today

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: done

src/components/simple_command_sequencer/gen/templates/command_sequences/name_commands.adb
cbrisebois-lasp: done

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: done

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: done

src/components/simple_command_sequencer/simple_command_sequencer.events.yaml
```diff
@@ -45,7 +45,8 @@ events:
     param_type: Command_Response.T
 
   - name: No_Frame_Available
-    description: A Run_Sequence command was received but all frames are in use
+    description: A Run_Sequence command was refused because every frame of the pool its sequence draws from (waiting or no-wait) is in use. The other pool may still hold idle frames; the two are kept apart on purpose so a late response to a finished no-wait run can never wake a waiting frame.
```
cbrisebois-lasp: is this too verbose?

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.ads
```diff
@@ -51,10 +51,14 @@ private
    type Instance is new Simple_Command_Sequencer.Base_Instance with record
       Sequence_Frames : Simple_Sequencer_Types.Sequence_Frame_Array_Access := null;
       Sequences : Simple_Sequencer_Types.Sequences_Access := null;
-      -- Where the next frame search starts: one past the frame claimed most
-      -- recently, so frames are handed out round robin and the frame released
-      -- most recently is the last one reused.
-      Next_Frame_Hint : Simple_Sequencer_Types.Frame_Id_Type := 0;
+      -- Frames 0 .. Num_Waiting_Frames - 1 serve sequences that wait for
+      -- sub-command responses; the frames after them serve no-wait sequences.
+      Num_Waiting_Frames : Simple_Sequencer_Types.Pool_Size_Type := 0;
```
cbrisebois-lasp: Why is there not a Num_No_Waiting_Frames variable?

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -18,34 +18,61 @@ package body Component.Simple_Command_Sequencer.Implementation is
...
 
    overriding procedure Init (Self : in out Instance; Config : in Simple_Sequencer_Types.Sequencer_Config) is
+      -- The model guarantees the two pools together make a valid frame count.
+      Num_Frames : constant Simple_Sequencer_Types.Num_Concurrent_Sequences_Type := Config.Num_Waiting_Frames + Config.Num_No_Wait_Frames;
    begin
-      Self.Sequence_Frames := new Simple_Sequencer_Types.Sequence_Frame_Array (0 .. Frame_Id_Type (Config.Num_Concurrent_Sequences - 1));
+      Self.Sequence_Frames := new Simple_Sequencer_Types.Sequence_Frame_Array (0 .. Frame_Id_Type (Num_Frames - 1));
       Self.Sequence_Frames.all := [for Id in Self.Sequence_Frames.all'Range => (Frame_Id => Id, others => <>)];
       Self.Sequences := Config.Sequences;
+      Self.Num_Waiting_Frames := Config.Num_Waiting_Frames;
```
cbrisebois-lasp: This is implemented a little strangly. Why do we not have a symmetric variable for Num_No_Waiting_Frames

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -18,34 +18,61 @@ package body Component.Simple_Command_Sequencer.Implementation is
    subtype Frame_Id_Type is Simple_Sequencer_Types.Frame_Id_Type;
 
    overriding procedure Init (Self : in out Instance; Config : in Simple_Sequencer_Types.Sequencer_Config) is
+      -- The model guarantees the two pools together make a valid frame count.
+      Num_Frames : constant Simple_Sequencer_Types.Num_Concurrent_Sequences_Type := Config.Num_Waiting_Frames + Config.Num_No_Wait_Frames;
    begin
-      Self.Sequence_Frames := new Simple_Sequencer_Types.Sequence_Frame_Array (0 .. Frame_Id_Type (Config.Num_Concurrent_Sequences - 1));
```
cbrisebois-lasp: Why do we need this? What walks all of the frames? I sorta feel like anything that walks all of the frames should be a function that accepts and argument about which pool it's walking.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -18,34 +18,61 @@ package body Component.Simple_Command_Sequencer.Implementation is
    subtype Frame_Id_Type is Simple_Sequencer_Types.Frame_Id_Type;
 
    overriding procedure Init (Self : in out Instance; Config : in Simple_Sequencer_Types.Sequencer_Config) is
+      -- The model guarantees the two pools together make a valid frame count.
+      Num_Frames : constant Simple_Sequencer_Types.Num_Concurrent_Sequences_Type := Config.Num_Waiting_Frames + Config.Num_No_Wait_Frames;
```
cbrisebois-lasp: why do we need this?

src/components/simple_command_sequencer/simple_command_sequencer.events.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Agreed. Every other event in the file is one sentence, and the rationale already lives in the generator doc's Frame Pools section and the config record's comment. Trimmed to the one fact an operator needs: which pool was full. Addressed in ebfa2d4.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.ads
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Agreed. The asymmetric field and the two hint fields are replaced by a `Frame_Pool_Range` record per pool (first, last, count, and the round-robin cursor) in an array indexed by the `Frame_Pool` enum, so `Self.Pools (Waiting)` and `Self.Pools (No_Wait)` are laid out and read identically. Addressed in ebfa2d4.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Same fix: Init now lays out both pools through one `Make_Pool` helper, and `Claim_Available_Sequence_Frame` takes the pool enum and works from that pool's record, so the only arithmetic left is the wrap of the cursor. Addressed in ebfa2d4.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
The total is needed for one thing, sizing the frame array. The routines that walk every frame do so because they are about frames as frames, not about allocation: routing a response to its frame by source id, assigning source ids at registration in id order, filling the summary packet in id order, waking parked frames on tick, the running-frame count, and Kill_All. None of them cares which pool a frame belongs to, so a pool argument would be one every caller passes as "both". Allocation is the only pool-aware operation, and after the fixup `Claim_Available_Sequence_Frame` is the only routine that takes a pool. No code change for this one beyond that boundary.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Two reasons, and the old comment only gave one. It sizes the allocation, and its subtype re-checks at Init the guarantee the model made at build time: the pool counts are modular, so a zero total inlined into the allocation would wrap to a huge value before the frame-id conversion rejected it, and the failure would point at the wrong line. The constant makes that check explicit; its comment now says so. Addressed in ebfa2d4.

src/components/simple_command_sequencer/types/simple_sequencer_types.ads
```diff
@@ -64,20 +64,31 @@ package Simple_Sequencer_Types is
...
 
-   -- Frames per sequencer instance. The upper bound is the most
-   -- Sequence_Frame_Summary entries that fit in one summary packet.
+   -- Frames per sequencer instance, both pools together. The upper bound is
+   -- the most Sequence_Frame_Summary entries that fit in one summary packet.
    subtype Num_Concurrent_Sequences_Type is Interfaces.Unsigned_32 range 1 .. Interfaces.Unsigned_32 (Packet_Types.Packet_Buffer_Type'Length / Sequence_Frame_Summary.Size_In_Bytes);
 
+   -- Size of one frame pool. A pool may be empty when no sequence draws from
+   -- it; the two pools together must make a valid Num_Concurrent_Sequences_Type.
+   subtype Pool_Size_Type is Interfaces.Unsigned_32 range 0 .. Num_Concurrent_Sequences_Type'Last;
```
cbrisebois-lasp, line 73: Is Pool_Size_Type ever telemetered? If so we shouldn't definitely make this smaller than Unsigned_32 otherwise it doesn't really matter, but should still probably be Unsigned_8?

src/components/simple_command_sequencer/types/sequence_frame_summary.record.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-description: Per-frame entry in the sequencer summary packet. One of these is produced for each sequence frame (Num_Concurrent_Sequences total) every time the summary packet is emitted.
+description: Per-frame entry in the sequencer summary packet. One of these is produced for each sequence frame (both pools, waiting frames first) every time the summary packet is emitted.
```
cbrisebois-lasp: reword - "...for each sequence frame in both pools every time..."

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
```diff
@@ -9,4 +9,8 @@ enums:
   - name: Sequence_Response_Behavior
     literals:
       - name: Send_After_Sequence_Start
-      - name: Send_After_Sequence_Completion
\ No newline at end of file
+      - name: Send_After_Sequence_Completion
+  - name: Frame_Pool
+    literals:
+      - name: Waiting
```
cbrisebois-lasp: - name: Wait
- name: No_Wait

this is more self consistent

src/components/simple_command_sequencer/gen/templates/command_sequences/name_summary_record.record.yaml
```diff
@@ -1,7 +1,7 @@
 ---
-description: Autocoded summary packet type for the {{ name }} command sequences suite. Contains one Sequence_Frame_Summary record per sequence frame ({{ num_concurrent_sequences }} total, from the suite's num_concurrent_sequences), in frame order.
+description: Autocoded summary packet type for the {{ name }} command sequences suite. Contains one Sequence_Frame_Summary record per sequence frame ({{ num_frames }} in all, the {{ num_waiting_frames }} waiting frames followed by the {{ num_no_wait_frames }} no-wait frames), in frame order.
```
cbrisebois-lasp: This should just be Wait_Frames and No_Wait_Frames

src/components/simple_command_sequencer/gen/templates/command_sequences/name.ads
```diff
@@ -107,10 +107,11 @@ package {{ name }} is
 
    Sequences : constant Sequences_Access := Sequences_Table'Access;
 
-   -- Passed to Simple_Command_Sequencer.Init. num_concurrent_sequences also
-   -- sizes the generated {{ name }}_Summary_Record packet type, so the frame
-   -- pool and the packet layout cannot disagree.
+   -- Passed to Simple_Command_Sequencer.Init. The two pool sizes together also
+   -- size the generated {{ name }}_Summary_Record packet type, so the frame
```
cbrisebois-lasp: This is worded strangely. "...together also size the generated..." This just doesn't make sense reword this comment

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -16,18 +16,30 @@ mapping:
     desc: |
       Custom preamble text to include in the generated Ada package.
 
-  num_concurrent_sequences:
+  num_waiting_frames:
```
cbrisebois-lasp, line 19: Same here this should just be num_wait_frames

src/components/simple_command_sequencer/types/simple_sequencer_types.ads
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Not telemetered: Pool_Size_Type appears only in Sequencer_Config, the compile-time constant the generated suite hands to Init, so its width has no runtime cost. On narrowing it: the cap it is bounded by is the packet buffer over the 8-byte summary entry, which is 125 for our 1004-byte buffer and 155 for the framework default, so Unsigned_8 would fit here but would silently reject a valid config on a project with a larger packet buffer. Unsigned_16 is the natural width (Frame_Id_Type is Unsigned_16 for the same reason), but that would mean narrowing the total type alongside it. Left as a subtype of the existing Unsigned_32 total type, per your call; no change.

src/components/simple_command_sequencer/types/sequence_frame_summary.record.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Reworded as suggested: "One of these is produced for each sequence frame in both pools, waiting frames first, every time the summary packet is emitted." Addressed in 91f110b.

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Renamed per the amended theme: the Frame_Pool literals are Waiting_Frame and Non_Waiting_Frame, and the same vocabulary now runs through the suite keys (num_waiting_frames / num_non_waiting_frames), the Config fields, the model, the templates, the tests, and the prose the pooling commit introduced. The word "no-wait" no longer appears in anything this commit added; the base commit's older comments were left alone to keep the diff to the pooling change. Addressed in 91f110b.

src/components/simple_command_sequencer/gen/templates/command_sequences/name_summary_record.record.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Covered by the rename: the template now reads num_waiting_frames and num_non_waiting_frames, matching the suite keys and the Config fields. Addressed in 91f110b.

src/components/simple_command_sequencer/gen/templates/command_sequences/name.ads
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Agreed, that sentence had the dependency backwards. It now says the suite's Summary_Record packet type is generated from the same two pool sizes, so the frame pools and the packet layout cannot disagree. Addressed in 91f110b.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Covered by the rename: the keys are num_waiting_frames and num_non_waiting_frames, and the model's rejection messages name them the same way. Addressed in 91f110b.

src/components/simple_command_sequencer/gen/models/command_sequences.py
```diff
@@ -130,6 +130,17 @@ def validate(self):
...
+        # A static arg is baked into the step table at build time, so it cannot
+        # carry a per-call value: any mention of Arg outside a pure reference
+        # would be emitted verbatim and fail to compile in the generated
+        # package, far from the yaml that caused it.
+        if self.arg and re.search(r"\bArg\b", self.arg):
+            raise ModelException(
+                f"Step {self.index} arg '{self.arg}' mixes a reference to the "
+                "sequence argument with other Ada. A step arg must be either "
+                "exactly 'Arg' or a field path 'Arg.Field[.Subfield]', or a "
+                "static Ada expression that does not mention Arg."
```
cbrisebois-lasp, line 142: add a sentence saying this feature is not *yet supported

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -359,6 +361,7 @@ package body Component.Simple_Command_Sequencer.Implementation is
                               (Sequence_Id => Frame.Sequence_Id, Frame_Id => Frame_To_Wake_Id,
                                Step => Frame.Step)));
                         else
+                           Frame.Step := @ + 1;
```
cbrisebois-lasp, line 364: wrap these three lines a wake_and_execute. call at other site as well.

src/components/simple_command_sequencer/simple_command_sequencer.simple_command_sequencer_commands.yaml
```diff
@@ -0,0 +1,20 @@
...
+description: |
+  Static commands for the Simple Command Sequencer component. Per-sequence
+  commands (one per declared sequence) are synthesised at assembly
+  load time by gen/models/simple_command_sequencer_commands.py.
+commands:
+  - name: Run_Sequence
+    description: Run a command sequence by ID. The synthesised per-sequence commands are the operator-friendly form; this is the underlying backbone they all dispatch through. Response behavior is the sequence's static configuration from the sequences model.
+    arg_type: Run_Sequence_Arg.T
+
+  - name: Kill_All_Sequences
```
cbrisebois-lasp, line 11: add to description to say that if you kill waiting_for_response sequences you enter a stray response unsafe state (add more details than this) don't use random jargon

src/components/simple_command_sequencer/simple_command_sequencer.simple_command_sequencer_commands.yaml
```diff
@@ -0,0 +1,20 @@
...
+    description: Halt every running sequence and return all frames to their initial state. Does not affect frames that were not running.
+
+  - name: Kill_Frame
```
cbrisebois-lasp, line 14: add to description to say that if you kill waiting_for_response sequences you enter a stray response unsafe state (add more details than this) don't use random jargon

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.ads
```diff
@@ -47,14 +48,26 @@ private
...
+   -- One frame pool: a contiguous run of frame ids and the round-robin cursor
+   -- for the next search in it. An empty pool has Count = 0 and is never
+   -- searched.
+   type Frame_Pool_Range is record
+      First : Simple_Sequencer_Types.Frame_Id_Type := 0;
+      Last : Simple_Sequencer_Types.Frame_Id_Type := 0;
+      Count : Natural := 0;
+      -- One past the frame claimed most recently, so frames are handed out
+      -- round robin and the frame released most recently is the last reused.
+      Next_Hint : Simple_Sequencer_Types.Frame_Id_Type := 0;
```
cbrisebois-lasp, line 60: change hint language, ugly

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
```diff
@@ -9,4 +9,8 @@ enums:
   - name: Sequence_Response_Behavior
     literals:
       - name: Send_After_Sequence_Start
-      - name: Send_After_Sequence_Completion
\ No newline at end of file
+      - name: Send_After_Sequence_Completion
+  - name: Frame_Pool
+    literals:
+      - name: Waiting_Frame
```
cbrisebois-lasp, line 15: call this waiting_for_response_frames

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
```diff
@@ -9,4 +9,8 @@ enums:
...
+      - name: Non_Waiting_Frame
```
cbrisebois-lasp, line 16: call this non_waiting_for_response_frames

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -16,18 +16,30 @@ mapping:
...
+    range: { min: 0 }
     desc: |
-      The number of sequence frames (concurrently running sequences) a
-      sequencer instance initialized with this suite allocates. This also
-      sizes the suite's generated summary packet ground type (one
-      Sequence_Frame_Summary per frame), so the frame pool and the packet
-      layout can never disagree. The upper bound -- how many frame
-      summaries fit in one packet buffer -- is enforced by the model
-      against the project configuration.
+      Frames for sequences that wait for their sub-commands' responses
```
cbrisebois-lasp, line 24: make these descriptions more symmetric

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -18,34 +18,58 @@ package body Component.Simple_Command_Sequencer.Implementation is
...
    overriding procedure Init (Self : in out Instance; Config : in Simple_Sequencer_Types.Sequencer_Config) is
+      -- Sizes the frame array, and re-checks at Init what the model guaranteed
+      -- at build time: a total outside 1 .. Num_Concurrent_Sequences_Type'Last
+      -- fails here, on the subtype, rather than as a wrapped modular value in
+      -- the allocation below.
+      Num_Frames : constant Simple_Sequencer_Types.Num_Concurrent_Sequences_Type := Config.Num_Waiting_Frames + Config.Num_Non_Waiting_Frames;
+
+      -- Lay one pool over Count frames starting at First, with its round-robin
+      -- cursor on its first frame. An empty pool is left at its defaults.
+      function Make_Pool (First : in Natural; Count : in Natural) return Frame_Pool_Range is
```
cbrisebois-lasp, line 29: consider renaming Frame_Pool_Range type to reflect that this record has more information than just the range.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -498,12 +522,15 @@ package body Component.Simple_Command_Sequencer.Implementation is
...
-            Self.Event_T_Send_If_Connected (Self.Events.No_Frame_Available (Time));
-            return Failure;
-         end if;
-         -- The next claim starts its search after this frame.
-         Self.Next_Frame_Hint := (if Available_Id = Self.Sequence_Frames.all'Last then Self.Sequence_Frames.all'First else Available_Id + 1);
+         declare
+            Pool_Kind : constant Sequence_Enums.Frame_Pool.E :=
+               (if Seq.Wait_For_Cmd_Resp then Sequence_Enums.Frame_Pool.Waiting_Frame else Sequence_Enums.Frame_Pool.Non_Waiting_Frame);
+         begin
+            if not Claim_Available_Sequence_Frame (Self, Pool_Kind, Available_Id) then
```
cbrisebois-lasp: rename back to find_available_sequence_frame

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -18,34 +18,58 @@ package body Component.Simple_Command_Sequencer.Implementation is
...
-      for Offset in 0 .. Count - 1 loop
+      for Offset in 0 .. Pool.Count - 1 loop
          declare
-            Id : constant Frame_Id_Type := Frames'First + Frame_Id_Type ((Natural (Self.Next_Frame_Hint - Frames'First) + Offset) mod Count);
+            Id : constant Frame_Id_Type := Pool.First + Frame_Id_Type ((Natural (Pool.Next_Hint - Pool.First) + Offset) mod Pool.Count);
          begin
             if Frames (Id).Status = Not_Running and then Frames (Id).Has_Source_Id then
                Frame_Id := Id;
+               -- The pool's next claim starts its search after this frame.
+               Pool.Next_Hint := (if Id = Pool.Last then Pool.First else Id + 1);
```
cbrisebois-lasp, line 66: perhaps move out of this if the name is find as apposed to claim? find implies no state manipulations, yet this line is a state change

@dinkelk you pick!

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.ads
```diff
@@ -47,14 +48,26 @@ private
...
+   end record;
```
dinkelk, line 61: This seems like a weird way to structure all this, and we are losing the Ada type system's natural index bounds checking. Why not just make two:
```
Sequence_Frames : Simple_Sequencer_Types.Sequence_Frame_Array_Access := null;
```
types, or similar. That way we can use the Ada type system without inventing our own bounds checking.
And in this way, the two frame pools are guaranteed to have no interaction because they are separate pieces of memory. Only thing that might be needed with this design might be a subprogram that given a source_id tells you which frame pool a command response belongs to.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -18,34 +18,58 @@ package body Component.Simple_Command_Sequencer.Implementation is
...
+      -- fails here, on the subtype, rather than as a wrapped modular value in
+      -- the allocation below.
+      Num_Frames : constant Simple_Sequencer_Types.Num_Concurrent_Sequences_Type := Config.Num_Waiting_Frames + Config.Num_Non_Waiting_Frames;
+
+      -- Lay one pool over Count frames starting at First, with its round-robin
+      -- cursor on its first frame. An empty pool is left at its defaults.
+      function Make_Pool (First : in Natural; Count : in Natural) return Frame_Pool_Range is
+      begin
+         if Count = 0 then
+            return (others => <>);
```
dinkelk: Not a fan of this return default language. Better to be explicit about what everything is set to for readability.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -16,18 +16,30 @@ mapping:
...
+    desc: |
+      Frames for sequences that do not wait for responses
+      (wait_for_command_completion false); they follow the waiting frames
+      in id order. The pools are separate so a response still in flight to
+      a finished non-waiting run can never wake a frame that is parked waiting.
+      Must be at least 1 if any sequence in the suite does not wait. The
+      two counts together size the suite's generated summary packet type
+      (one Sequence_Frame_Summary per frame); their sum must be at least 1
+      and at most the number of summaries that fit one packet buffer,
+      enforced by the model.
```
dinkelk: Some checks might be good here in the model. Like this should NOT be >0 if there are no non_waiting sequences, and vise versa for num_waiting_frames. And these SHOULD be >0 if there is at least 1 sequence of the designated type. And to go with that, these should default to zero if not specified.

dinkelk (changes requested):

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.ads
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Renamed to `Next_Search_Start`, and retyped along with it: after discussing this with Connor, we dropped the range-over-a-shared-array record entirely. Each pool now owns its own frame array, so the cursor is an index into that array rather than a global frame id that only the claim arithmetic kept in range. Addressed in 9a381ea.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.ads
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**

Adopted. Each pool is a `Frame_Pool` record holding its own `Sequence_Frame_Array_Access` and round-robin index, with the non-waiting-for-response array starting where the waiting-for-response one ends, so frame ids stay global and contiguous. That keeps Kill_Frame, the summary packet, and the tests unchanged, while the index subtype bounds every access and the idle-frame search walks one pool's own range without modular arithmetic. The search is a pure `in` function; Run_Sequence advances the pool's cursor when it seeds the frame it took. The source-id lookup returns the pool along with the frame id.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**

`Make_Pool` and its `others => <>` return are gone with the range record; `Init_Pool` in Init now builds each pool with every field named, and an empty pool is an explicit empty array rather than a defaulted record.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Done in full. Each pool is checked both ways in the model: at least one frame when the suite has a sequence of that kind, and exactly zero when it has none, since nothing could ever claim those frames. Both keys (`num_waiting_for_response_frames` / `num_non_waiting_for_response_frames`) are optional and default to 0.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Rewritten alongside the model checks so both keys state the same rule in the same words: what the frames are for, which ids they take, the at-least-one and exactly-zero conditions, and the default. The pool-separation rationale and the packet-sizing note now sit once, after the two definitions. Addressed in 0802cf3.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Added `Wake_And_Execute` next to `Execute_Sequence`: it steps the frame past the command or sleep it was parked on, marks it Running, and runs it to its next park or the end of the sequence. Both wake sites, the response handler and the tick handler's expired sleep, now call it; the timeout and abort paths are left alone since they finish rather than wake. Addressed in 7256b98.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Renamed, and carried through everything that used the old word: the enum literals, the suite keys (`num_waiting_for_response_frames` / `num_non_waiting_for_response_frames`), the Config fields, the model, the templates, the tests, and the prose. Connor's reasoning: a frame in the second pool can still park on a sleep step, so "non-waiting" was misleading; what the pools actually separate is whether a frame ever waits for a command response. Addressed in 42520a6.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Same rename; see the thread on the first literal for the full scope. This one is `Non_Waiting_For_Response_Frame`, with `num_non_waiting_for_response_frames` as its suite key. Addressed in 42520a6.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/gen/models/command_sequences.py
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Added the sentence: the message now says that mixing literals and sequence argument fields in one step arg is not yet supported, before listing the two forms that work today. The code comment above it was reworded in the same spirit, so it describes a generator limit rather than a rule. Addressed in 9059f2e.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/simple_command_sequencer.simple_command_sequencer_commands.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Kill_All_Sequences now spells out the consequence in plain words: killing a sequence that is waiting for a command response does not cancel that command, so its response can arrive after the frame has been released and claimed by a new sequence. It states what the sequencer does with it: a response whose command id does not match the new sequence's pending step is ignored, one with a matching id would wake that step early. Addressed in f65032b.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/simple_command_sequencer.simple_command_sequencer_commands.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Same text as Kill_All_Sequences, phrased for a single frame: killing a frame that is waiting for a command response does not cancel that command, its response can still reach the frame after a new sequence has claimed it, and only a matching command id would wake the new step early. Addressed in f65032b.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Renamed as part of the restructure in 9a381ea: the record is now `Frame_Pool`, since it owns the pool's frame array and its search cursor rather than describing a range over a shared one. `Frame_Pool_Range` no longer exists.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Renamed back to `Find_Available_Sequence_Frame`, and made the name true: the function now takes `Self : in Instance`, so the compiler guarantees the search changes nothing. The round-robin cursor advance moved into Run_Sequence, next to the record seed, which is the one place the frame is actually taken. Behavior is identical and the tests are unchanged. Addressed in cd426eb.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Resolved by moving the state change out rather than by picking a name: the cursor write now lives in Run_Sequence where the frame is seeded, and the search is a pure `in` function. Connor and I judged "Claim" the worse option, since the routine never marked the frame taken (Run_Sequence overwrites the whole record right after), so the name would have promised a claim it did not make. Addressed in cd426eb.

Response reviewed and approved by @cbrisebois-lasp.

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
```diff
@@ -9,4 +9,8 @@ enums:
   - name: Sequence_Response_Behavior
     literals:
       - name: Send_After_Sequence_Start
-      - name: Send_After_Sequence_Completion
\ No newline at end of file
+      - name: Send_After_Sequence_Completion
+  - name: Frame_Pool
```
cbrisebois-lasp, line 13: add descriptions about pools

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
```diff
@@ -9,4 +9,8 @@ enums:
...
+    literals:
+      - name: Waiting_For_Response_Frame
```
cbrisebois-lasp, line 16: add description here

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
```diff
@@ -9,4 +9,8 @@ enums:
...
+      - name: Non_Waiting_For_Response_Frame
```
cbrisebois-lasp, line 18: add description here

cbrisebois-lasp: done

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
cbrisebois-lasp: done

src/components/simple_command_sequencer/types/sequence_enums.enums.yaml
cbrisebois-lasp: done

dinkelk (approved):

PR #230, opened 2026-09-17
Simple command sequencer patches

cbrisebois-lasp:


src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
```diff
@@ -99,6 +99,11 @@ mapping:
               If true, the sequence will continue executing even if a sub-command
               fails. If false, the sequence will abort on the first failure.
 
+              A sub-command failure is only visible when the sequence waits for
```
cbrisebois-lasp, line 101: do other keys describe there failure cases as well? I'm wondering if this is noise?

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -122,7 +125,9 @@ package body Component.Simple_Command_Sequencer.Implementation is
          Self.Data_Product_T_Send_If_Connected (Self.Data_Products.Sequences_Failed_Count (Time, (Value => Self.Sequences_Failed_Count)));
          Self.Data_Product_T_Send_If_Connected (Self.Data_Products.Last_Sequence_Failed (Time, (Value => Frame.Sequence_Id)));
       end if;
-      Send_Frame_Count_Data_Products (Self, Time);
+      if Update_Frame_Counts then
+         Send_Frame_Count_Data_Products (Self, Time);
```
cbrisebois-lasp, line 139: why is this called in Finish_Sequences? Shouldn't it just be called in Kill_All_Sequences?

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
```diff
@@ -529,14 +534,19 @@ package body Component.Simple_Command_Sequencer.Implementation is
...
          if Frame.Status /= Not_Running then
             -- Any deferred reply is sent now with Failure, so the operator's
             -- command does not hang.
-            Finish_Sequence (Self, Frame, Command_Response_Status.Failure, Time);
+            Finish_Sequence (Self, Frame, Command_Response_Status.Failure, Time, Update_Frame_Counts => False);
+            Killed_Any := True;
          end if;
       end loop;
+      if Killed_Any then
+         Send_Frame_Count_Data_Products (Self, Time);
```
cbrisebois-lasp, line 560: Is this not a duplicate of what's already in Finish_Sequences?

src/components/simple_command_sequencer/gen/schemas/command_sequences.yaml
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Yes, the schema states its rules inline as a convention: `num_concurrent_sequences` says "enforced by the model", both timeouts say "bounded so the count fits a Time_Span", `name` says "must follow Ada naming conventions", the step form says "mutually exclusive; exactly one must be present", and the trailing comment block lists the model-enforced rules. This one is the longest of them because it also gives the reason, but it is the same kind of statement. Kept as is per your call; no change.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Finish_Sequence is the single exit path every run goes through; it has ten callers (completion, abort on a failed sub-command, response timeout, out-of-range sleep and timeout, invalid dynamic argument and sleep, Kill_Frame, Kill_All, and the deferred-claim failure). Frame_Running_Count and its high-water mark have to update every time a frame goes idle, so the recount lives in the one place all ten paths share rather than at nine separate call sites. Kill_All is the only caller that idles several frames in one command, so it alone opts out per frame via the defaulted flag, the same shape as Is_Load on the CCSDS parameter table router's private helpers. No change.

src/components/simple_command_sequencer/component-simple_command_sequencer-implementation.adb
cbrisebois-lasp: **[Agent reply — Claude Fable 5.1]**
Not a duplicate: this is the single recount that replaces the per-frame ones Kill_All suppresses with Update_Frame_Counts => False. Before this commit a kill of N frames emitted the two frame-count data products N times with intermediate counts; now they go out once with the final count. No change.

